PR #42, opened 2018-02-12
Change of link

SamCosby:
Link under 'Modern Authentication Failure' currently fails with a 404. Changed it to be appropriate.


LolaJacobsen:
I made this fix manually (your URL contained "en-us" which we don't publish). It'll go live in a few minutes. Thanks for raising this PR!

PR #45, opened 2018-02-15
Update Call Quality Dashboard's "Is Teams" dimension for interop scenarios

GageAmes:
In interop scenarios, the "Is Teams" dimension will have "True" as the value.


tonysmit:
Approved and republished live. Thanks Gage!

PR #47, opened 2018-02-16
Romanma working

RomanMSFT:
stub out and edit of Guest Checklist


tonysmit:
Lola. I cancelled this. I was helping Roman in CSS and we were having issues. We have a work around so closing this.

PR #49, opened 2018-02-21
Corrected link to the spreadsheet with countries and regions 

robdy:
Looks like the link is no longer correct (file moved), I corrected it.


tonysmit:
@robdy Thank you for fixing this!

PR #96, opened 2018-04-04
Updated Surface Pro Drivers links

aageorge:
Links now point to download for all driver versions as required by setup script.


jambirk:
This update should be rejected, or at least not accepted without further discussion. The FWlinks shown here in the requirements page are also in the console creation page, and they are both deliberately pointing directly to the driver packages rather than the download details for the packages.

jambirk:
we've changed the redirect target, so changing the link on the page is no longer necessary.

PR #107, opened 2018-04-09
Update msi-deployment.md

RomanMSFT:
Added clarification to the "Important" note section


tonysmit:
Thanks Roman!! I picking up for Lola.

PR #110, opened 2018-04-10
Correct the type of "Is Server" dimensions in CQD

GageAmes:
The "Is Server" dimensions in CQD are string enumeration values, not Boolean.


tonysmit:
Thanks Gage!

PR #111, opened 2018-04-10
Update topology.md

Cloud-Writer:
In the first sentence of the last paragraph in this article, please delete "and using". This is misleading. Persistent Chat performance counters count a user as an "active user" as soon as they are logged in to their client and activated for PChat. A client app will also start sending SIP messages as soon as it is logged in, generating load.

SAP tried to do their capacity planning based on how many users would be actively using persistent chat at any given time, which would be significantly lower than how many users would be activated and logged in at the same time. Thus, their environment was severely undersized. This caused them a critical service outage affecting other Lync services as well and led to an escalation.


jambirk:
discussed this change with Meera. Side note: she was forced to use chrome, IE was not allowing her to submit a pull request, or to toggle beween edit and preview mode

PR #126, opened 2018-04-18
Update syntax and related note for set-csuser

carolynblanding:
Set-CsUser requires providing User Principal Name for the -Identity. No other options are supported at Set-CsUser will fail with Object reference not set to an instance of an object.


carolynblanding:
Hold off on this request. Looks like we may have a bug here. Per cmdlet help, Sip Address, UPN, or Active Directory Display Name should work. Doesn't work for me but Jens reports that it works for him.

carolynblanding:
PS - I double checked and it seems that the problem is only with the Display name. Sip address and UPN are both working. Raising for investigation. Again, hold off on this change. I think the docs are find and we have code issue.

PR #129, opened 2018-04-18
Update reset-a-conference-id-for-a-user.md

cowitkow:
Per the SFB product group, changes to static are no longer supported. Engaging Support will trigger the same response.


tonysmit:
Thank you! Appreciate it. Tony

PR #130, opened 2018-04-18
All scenarios of 'Static' need to be taken out.

SamCosby:
Changing document to state that Static is no longer supported in any way, shape or form, per PG.


tonysmit:
Hey Sam. I'm working on removing the static IDs out with a PM. I'm waiting to hear back from him so I can scrub some other topics for this also. I'm listening and thanks for the fix! Appreciate it. Tony

PR #145, opened 2018-04-25
Metadata Edit as per VSO: BUG:2283348

v-srteeg:


tonysmit:
@v-srteeg - Hi. Why are you changing this?

dstrome:
@tonysmit , these changes move the SCS metadata from ms.custom to ms.collection (where needed) and change loc priority for any SCS topic to "Priority" to force a human translation review. Please review the conflict and make sure the correct content is merged along with the updated metadata.

PR #147, opened 2018-04-26
Tidy up PowerShell script for Teams clean up

ChrisLGardner:
Use full parameter names where possible to ensure readability and maintainability by many users. 
Use Test-Path instead of [System.Io.Path], again for readability. Leave the Combine methods in place as they are more readable than the alternative approach of nested Join-Path calls.
Change Write-Output in the catch block to Write-Error. Write-Output should only be used to pass things along the pipeline, for other information output there are a number of other streams for use and Write-Error is the most applicable here. There is a case for using Write-Warning instead for user interaction purposes, since many users are less likely to read the red text that Write-Error defaults to, but that doesn't quite fit with what the cmdlet does as the information it would be relaying is an error and not a warning.


tonysmit:
@ninadara - Can you verify this or is there someone that owns the script? I don't want to merge it until someone checks off on it. Thanks, Tony

PR #150, opened 2018-04-26
Clarified the scenario

karuanag:
This is can't be used to migrate SFB meetings to Teams. That functionality doesn't exist.  This wasn't clear to the documentation audience who may be looking for that function.


tonysmit:
I also removed it from the Teams TOC.

PR #203, opened 2018-05-16
FIxed HA Terminology

allanhirt:
AlwaysOn has had a space for 3 years. It hasn't been fixed in all UI elements, and I di didn't fix those. Most, if not all, SQL Server documentation has been changed to reflect this. Yours should as well, hence these changes. As per
https://docs.microsoft.com/en-us/sql/database-engine/availability-groups/windows/always-on-availability-groups-sql-server?view=sql-server-2017
"[NOTE] Always On availability groups is the full, formal name for this availability feature. The abbreviation is AG, not AOAG or AAG."

Also, Failover Clustering is the feature, a Windows Server Failover Cluster (WSFC) is the cluster itself.

I also deleted this
> [!IMPORTANT]
> Instance names for multiple Always On avialbility group instances must be the same. 

This is not true. INstance names can be different. The data paths should be the same, and saying multiple Always On AG instances makes no sense in the SQL Server world.

Is this in your UI? If so it needs to be fixed:
    
    - In the **Define New SQL Store** page, first select the **High Availability Settings** check box, and then make sure that SQL AlwaysOn Availability Groups appears in the drop-down box.

### DIFF
--- a/Skype/SfbServer/deploy/deploy-high-availability-and-disaster-recovery/alwayson-availability-group.md
+++ b/Skype/SfbServer/deploy/deploy-high-availability-and-disaster-recovery/alwayson-availability-group.md
@@ -1,5 +1,5 @@
 ---
-title: "Deploy an AlwaysOn Availability Group on a Back End Server in Skype for Business Server 2015"
+title: "Deploy an Always On Availability Group on a Back End Server in Skype for Business Server 2015"
 ms.author: heidip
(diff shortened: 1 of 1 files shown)


kenwith:
@allanhirt - thank you for this PR and taking the time to correct this documentation. I am also raising this to the team Skype for Business Server team regarding the UI issue you brought up. Really appreciate it!

PR #214, opened 2018-05-18
Revised failure message wording

greiginsydney:
The vanilla "Multiple Editions" ISO of Windows does not include Enterprise Edition and (correctly) fails the version test. Amended the error message to add this as a possible reason for the failure.


jambirk:
After conferring with David Groom, we're shutting down this page completely and keeping the canonical scripts on Download Center. Will redirect pointers to this page in a separate PR

PR #226, opened 2018-05-21
Cleaning unnecessary spacing

nsshommo:
Unnecessary spacing is cleaned with this PR.


jambirk:
I saw this too and wanted the relevant pm to comment before making the pull.

James

kenwith:
@jambirk - the only change was removing white space. Not sure what the PM would comment on.

jambirk:
Not sure either, but he's notified in case we ever need to revert.

PR #276, opened 2018-05-29
Updated links

nsshommo:
Link to the planning tool is updated.

### DIFF
--- a/Skype/SfbServer/management-tools/planning-tool/planning-tool.md
+++ b/Skype/SfbServer/management-tools/planning-tool/planning-tool.md
@@ -17,7 +17,7 @@ description: "Guidance on using the Skype for Business Server 2015 Planning Tool
  
 Guidance on using the Skype for Business Server 2015 Planning Tool.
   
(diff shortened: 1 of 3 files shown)


nsshommo:
@jambirk Can you merge it?

jambirk:
Nassim, the extra padding in the See Also sections is being handled in bulk. The changes to the link you're requesting are not adequately explained: why is the  change more correct than what's already there?

nsshommo:
@jambirk Thanks for your reply. From now, I will not edit the extra padding in the See Also section.

The present link goes to "Microsoft Lync Server 2013, Planning Tool" whereas it should go to "Skype for Business 2015, Planning Tool". That's why I have changed the link from "https://go.microsoft.com/fwlink/p/?LinkID=282725" to "https://www.microsoft.com/en-us/download/details.aspx?id=50357".

jambirk:
in which case the correct action is to change the fwlink's target rather than hard code in another link. It's quite likely doing so will fix similar issues in other files. I've done so now that I understand what the issue was

PR #301, opened 2018-06-04
Update video-based-screen-sharing.md

brandber:
Corrected a RDP table reference noted as RPD.


kenwith:
@brandber - great catch, thank you for updating! Appreciated!

PR #303, opened 2018-06-04
Update to include new tile to display OS version distribution and fix…

Turgayo:
… couple of typos(missing v2s)

### DIFF
--- a/Skype/SfbServer/deploy/deploy-clients/with-oms.md
+++ b/Skype/SfbServer/deploy/deploy-clients/with-oms.md
@@ -28,22 +28,22 @@ By following this guide, you can use a dashboard like the following example to g
 At a high level, you need to perform the following tasks:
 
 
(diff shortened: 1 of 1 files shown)


jambirk:
Please hold this PR, we need to sync this update's timing  with a change in the SRS installer later this month.

Turgayo:
This update has nothing to do with the SRS installer change, but if you want to release both changes at the same time, that's OK. You can merge this update when the timing is right.

jambirk:
We generally wait for the installer to update, and then make changes to the other scripts or assets as part of the same upload process by release ops. It's not as responsive a process as GitHub, but the SRS  installer has updated every month or two since we launched.

kenwith:
@jambirk and @Turgayo - can we merge this yet?

jambirk:
The release supporting Teams was released to the MS store, so the version on download center should get posted this week.  I'll do the merge when it's available.

PR #445, opened 2018-06-14
Add note about limit.

kenwith:


kenwith:
@ninadara - wasn't positive on this one but added what I think. CC @chuckedmonson

kenwith:
Sorry, I was confused. This is not valid. I was reading something about Azure. Closing this one.

PR #493, opened 2018-07-03
Configure online hybrid med server - Change

savusca:
Please note that it is not possible to run the Set-CsHybridMediationServer cmdlet against a licensed account. Following error will be thrown:
"User with identity "user" has licenses assigned. It cannot be set as Hybrid Mediation Server. Please remove all the licenses for it or re-create a new user without any license and try again."

However, license is still required to be added on this account for it to be replicated into the SfB online directory. 
Correct action would be:
1. Create user and assign license. 
2. Wait up to one hour for the account to be replicated into the SfB online directory.
3. Remove the license.
4. Might take 5-10 minutes after license removal. (time it takes for the customer to connect to remote ps, should be enough).
5. Run the set-csHybridMediationServer cmdlet.

Please let me know if there are any questions.
Thanks,
Sabin


MichaelLaMontagne:
You will need to run the set-csHybridMediationServer cmdlet again after the 30 day SFBO license deprovisioning process kicks in and clears the FQDN property.

PR #540, opened 2018-07-12
Revert VBSS metric name change and remove extra table column

GageAmes:
The VBSS classifier uses the same metrics as video streams, just with different thresholds. This is because VBSS streams are really just video streams as far as metric reporting is concerned.

I've also removed an erroneous extra "unclassified" table column in the VBSS table's Video Post FECPLR row.


kenwith:
@GageAmes - thanks for this. Changing to "Video" makes it much clearer. Appreciated!

PR #573, opened 2018-07-20
Update xml-config-file.md

gregbari:
Updated XML settings content for Teams Meetings enablement, Teams as default calling client, and Proximity Beacon support.


davgroom:
@kenwith - please proceed.

PR #574, opened 2018-07-20
Update plan-hybrid-connectivity.md

juancal:
Added an important requeriment for the hybrid on the DNS section


juancal:
Hi Carolyn. I agree the phrase "Depending on how DNS is configured in your organization, you may need to add these records to the internal hosted DNS zone for the corresponding SIP domain(s) to provide internal DNS resolution to these records." says it all. But, this being not a requirement for federation, the sentence can be easily neglected. I think we should either highlight the phrase with a "Note:" or "Important:" header, or mention explicitly that DNS resolution of the SRV record by the Edge server is required for the hybrid configuration to work.

PR #595, opened 2018-07-27
Update reference-topologies.md

martinboam:
added spaces to **Exchange UM deployment.**
-  with bothon-premises deployments
-  andhosted Exchange UM.


kenwith:
@martinboam - thanks for catch and the PR. Much appreciated!

PR #608, opened 2018-07-30
Fix name of Office SKU

DHB-MSFT:
Fix to correct naming.


kenwith:
@DHB-MSFT - thank you for catch. Much appreciated!

PR #610, opened 2018-07-31
Fixed table

YuriySamorodov:


kenwith:
Got this, thank you @YuriySamorodov for catch. Much appreciated! Should be live shortly.

PR #627, opened 2018-08-02
Update upgrade-and-coexistence-of-skypeforbusiness-and-teams.md

revaldiv:
At Line 48 (table for Islands - advantages column) added " *" .  
At Line 54 included language to denote the starred item in table: * Except in the case where chat is initiated by a Teams user, the chat will be received in Skype for Business provided the recipient is homed online and never logged in to Teams.

Adding the language so elimiate confusion as we do include here: https://docs.microsoft.com/en-us/MicrosoftTeams/migration-interop-guidance-for-teams-with-skype#detailed-mode-descriptions


LolaJacobsen:
I added @dearbeen as a reviewer. I'd like to have her approve this PR as well

kenwith:
@dearbeen - can you please let us know if this can be merged?

revaldiv:
@kenwith @LolaJacobsen @dearbeen 

I believe things have changed so this may not be the case anymore.  Seems like there is some confusion still regarding Legacy and Islands and perhaps the other document may need to be changed instead as it states this is possible in Islands when perhaps it shouldn't.  @bjwhalen might be able to comment on this.

kenwith:
@revaldiv - closing this but please re-open if you talk with @bjwhalen on how to handle.

PR #629, opened 2018-08-02
Use Markdown

anthonypants:
At the very least, when the document states "*skypespaces*, including asterisks" you should make sure that the asterisks are preserved.


kenwith:
@anthonypants - much appreciated! This probably got mangled in the conversion from TechNet. Thank you very much for catching and fixing!

PR #631, opened 2018-08-03
update tenant throttling numbers

amareshbAtMicrosoft:
we have increased the tenant throttling to 20, so updated the description above


kenwith:
@amareshbAtMicrosoft - thanks for update!

PR #644, opened 2018-08-06
removed empty links that don't go anywhere

sonu-arora:


kenwith:
@sonu-arora - thanks for the help on this and taking the time to submit a PR! Much appreciated. Let's ask article owner about those links and what to do with them too.

@LolaJacobsen - this PR removes a link to 3rd party terms.. maybe we should find the link to put in though instead of removing it? What do you think?

tonysmit:
Closing this as I rewrote the content and fixe all of the links that were broken.

PR #665, opened 2018-08-13
Editing page to be current with R4 settings

SamCosby:
Currently, R4 (General Ring) customers do not have TeamsOnly parameters, nor do they have many of the parameters listed in this page. Currently updating the page to be in-tune with what's actually available to Public Facing Tenants.


SamCosby:
We should recommend Admins use Grant-CsTeamsUpgradePolicy if they want to enable/flight TeamsOnly users for now until this rolls out to GA.

kenwith:
@SamCosby - I wonder if rather than remove entirely we should add a note that says this is not rolled out to everyone? What about the people that are able to see these params? @LolaJacobsen FYI

kenwith:
@islubin - can you provide guidance here? what should be visible on this page?

islubin:
@bjwhalen should be the POC for updates on this page!  @bjwhalen can you comment on Sam/Ken's suggestions?

kenwith:
@SamCosby - maybe a note that these are not available to everyone and roll out in rings? I don't think we can remove them but haven't been able to touch base with @bjwhalen or @islubin on this one.

PR #674, opened 2018-08-16
Licensing requirements; SDS sets hidden membership

gasperj:
Office 365 Groups Naming Policy was added to the AAD_Basic_EDU license.
https://docs.microsoft.com/en-us/azure/active-directory/users-groups-roles/groups-naming-policy

Office 365 Groups Expiration Policy requires Azure AD Premium license.
https://docs.microsoft.com/en-us/azure/active-directory/users-groups-roles/groups-lifecycle

SDS creates groups with hidden group membership from non-members, required to meet FERPA privacy policy. It's important to call out doing the same for non-SDS created class groups in an education environment.


kenwith:
@gasperj - thanks! much appreciated!

PR #694, opened 2018-08-22
update for Public team list not displaying

ChelseaStats:
Workaround is ineffective if you don't know what team you should be searching for, this needs to be communicate by the team owner through some other medium. Not practical for searchers to go through each letter of the alphabet hoping to find a match, for example.

### DIFF
--- a/Teams/Known-issues.md
+++ b/Teams/Known-issues.md
@@ -210,7 +210,7 @@ This article lists the known issues for Microsoft Teams, by feature area.
 
 |**Issue title**|**Behavior / Symptom**|**Known workaround**|**Discovery date**|
 |:-----|:-----|:-----|:-----|
-|Public team list does not display all teams  <br/> |The list of public teams is based on the Microsoft Graph.  <br/> |If you don't see a team, try searching for it in the top right search box and team owners communicate team names to colleagues <br/> | 7/21/17  <br/>|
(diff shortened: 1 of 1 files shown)


kenwith:
@ChelseaStats - thanks for adding the value on this from your expertise. Maybe change it to: "try searching for it in the top right search box. Also, the team owners should communicate team names to colleagues since many teams could show up in the search results." ?

ChelseaStats:
Thanks for the update. I'm not sure this helps the situation however. 
It is not multiple teams might appear in the search (if that is the case it would be easy to see which one they wanted to join), It is the issue is there is no option to see all public teams including those that used to be private (the current bug). 

When searching 'a' no teams are shown
When searching 'b' no teams are shown
...
when searching 'm' - MicrosoftBugFixing appears (for example).

that is a lot of searching to find the one team.

Does that make sense?

islubin:
@ChelseaStats I'm a little confused as well.  Are you saying that when you go to the "Join or create a team" page, you don't see all teams in the organization - both public and private?  Or that the search isn't returning teams when you look for them?

ChelseaStats:
**I am not seeing any teams in my org.** 

We had around 30 developers in my org that couldn't see a team we had set up privately among a small group doing a test - we then wanted to roll this out to all staff but they couldn't see the team. we tried sending the link - but that didn't work either. We then found a work around by asking them to search for the team and then it appeared.

 (I believe the original bug was any team that was private and had since been made public was not showing)

I hope this helps.

kenwith:
@ChelseaStats - going to merge this since it will be beneficial even if there is a bug that gets resolved.

@islubin - Sounds like the possible bug is "I believe the original bug was any team that was private and had since been made public was not showing".

PR #719, opened 2018-08-30
Update pilot-essentials.md

carolynblanding:
In light of new Upgrade Modes, we need to expand on guidance around upgrading users to Teams. Islands Mode only works as long as all users using both SfB and Teams side by side. Once any user is upgraded to Teams there is a danger that messages to SfB users who have never used Teams will fail to send and there will be no indication otherwise because these users will look like Teams users since drawbridge has been removed. We should also provide guidance to IT Pro's that they can view user activity/last log in times for Teams in the Modern Portal. 

Legacy behavior. Teams user initiates chat to SfB user who has never signed into Teams. Teams shows this user as an SfB user with option to Invite them to Teams. 

New behavior. Teams user initiates chat to SfB user who has never signed into Teams and has Islands Mode assigned. Teams shows this chat as a regular Teams chat and messages will route to Teams chat service. The end user will not be aware that the recipient has never signed into Teams and won't receive the Chat. 

Please reach out to cbland@microsoft and kkodali@microsoft.com for more information if needed.


kenwith:
Closing this PR since it is more of a note instead of doc change.

PR #744, opened 2018-09-07
remove links to "phones and devices home" and "compatible devices pro…

dougand-MSFT:
…gram"

these links are redundant given the current tree structure of the site, and one of them was an incorrect link since it pointed to the Lync variant of the page instead of the SfB variant.


jambirk:
@dougand-MSFT I don't agree with the edit, though I might if the comment/request was explained better.

jambirk:
The page has validity for Lync and SfB, it's not just one or the other, so it seems appropriate to point to both in the links.

PR #758, opened 2018-09-13
Fixed typo causing table pipes to show

karuanag:
Fixed video display & typo


kenwith:
@karuanag - it looks like there is a conflict. Please resolve.

PR #761, opened 2018-09-14
Added resolution for issue where the CloudConnector module cannot be …

savusca:
…imported


savusca:
> @savusca - is this an issue in the known issues list? If so, it would be better to add a link than duplicate the content?

It is not in the known issues list (as far as i know of). If you saw this in the known issues list, can you please share the link?

kenwith:
@savusca - I didn't see it in known issues either. Looks good, thanks for making the docs better!

PR #766, opened 2018-09-14
Update Known-issues.md

cowitkow:
Filing against bug - Fix would need to occur against https://domoreexp.visualstudio.com/MSTeams/_workitems/edit/287249


cowitkow:
@marclauinger - Can you comment on the above? Trying to get this issue added as we're seeing a number of customers raising this in the past couple months

marclauinger:
This looks good.  Go ahead and publish.

PR #778, opened 2018-09-19
Remove script reference

MattWade-MS:
Cxdcalldata script was removed from powershell gallery by author and is no longer available for general use.  Removed references from guide.


MattWade-MS:
Any updates here?  With the source script no longer available I want to get this updated so customers don't get sent to an empty script page.  :)

PR #785, opened 2018-09-20
updated line 345

AndreaBarr:
Changed command New-CsOnlineVoiceRoute to Get-CsOnlineVoiceRoute so that it correctly correlates with line 342


AndreaBarr:
@nmurav Ken wanted to make sure you were kept in the loop for this update.

PR #821, opened 2018-09-27
Corrected Set-UserPhoto Example

Windos:
`-Preview` was between `-PictureData` and the photo variable making the command invalid.


kenwith:
@Windos - thanks for fix and making the docs better. Much appreciated!

PR #823, opened 2018-09-27
Update validated-lync-apps.md

mregaya:
the applications are certified for Lync and not skype for business server. The "applied to" field creates confusion for customers, insisting that the documentation clearly shows that this particular list of certified application is also certified for skype for business.


jambirk:
while this comment makes a lot of sense and I've approved the update, we should have a short discussion with the PM reviewer for this topic area. I have a feeling we will need to make similar changes to a number of other topics. Please email me off GitHub, my MSFT alias is the same as it is on GitHub.

PR #849, opened 2018-10-08
Update guest-access-checklist.md

MarcoCarampanta:
Suggestion to change the article as stated above, since the Slider in the currently "Live" version of the article has been deprecated and should not be used currently (since 2nd of August)


LolaJacobsen:
Closing/rejecting because Somak is incorporating these changes into an overhaul of the Guest Access docset.

PR #851, opened 2018-10-08
typo + note

sohailta:


sohailta:
fixed a typo and added note for TLS 1.2 support which is a general question we are getting after customers read this article. Please merge to live after checking/

PR #856, opened 2018-10-09
Update prerequisites.md

geetha-gangaiah:
Fix broken link for Microsoft .NET Framework 4.5  download

### DIFF
--- a/Lync/LyncTools/prerequisites.md
+++ b/Lync/LyncTools/prerequisites.md
@@ -54,7 +54,7 @@ To run the Lync Server 2013 Stress and Performance Tool on your Lync Server 2013
 
 Your client computer must meet the following software requirements:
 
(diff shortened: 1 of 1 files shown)


jambirk:
This change was not approved for multiple reasons. The URL was changed, but the link text was not and now they are out of sync. Also, exchanging one fwlink for another is not proper procedure, the FWLink target itself needs to be updated and the fwlink in the article stays the same. Finally, the replacement fwlink is now language and country specific, things like en-us or en-in should be trimmed out of public facing fwlinks and the user's system should be allowed to freely choose the loc properties.

jambirk:
Looking at this more closely, there is still a definite issue with the link. Text mentions .NET 4.5 but the link target is .NET 3.5.

jambirk:
okay. Verified that there are no other instances of the fwlink. Updated the fwlink metadata to point to the .NET 4.5 download. there should be no further need for this change.

PR #862, opened 2018-10-10
Update set-up-guests.md

expiscornovus:
Updated the article so it uses the new Teams & Skype for Business admin center interface.


LolaJacobsen:
Rejecting because Somak is incorporating these changes into a current Guest Access documentation overhaul.

PR #875, opened 2018-10-11
Needs updated.

SamCosby:
Update required for new Guest Access enablement.


LolaJacobsen:
Rejecting this PR because Somak will make these updates in the overhaul of Guest Access documentation that's happening now.

PR #896, opened 2018-10-13
Updated -ReplaceExistingSettings switch

Atreidae:
Hi all, First PR here. 
the -ReplaceExistingSettings switch should be -ReplaceExistingRGSSettings as per https://docs.microsoft.com/en-us/powershell/module/skype/Import-CsRgsConfiguration?view=skype-ps

Tested on a prod environment with Skype and Lync 2013 and confirmed the same.


kenwith:
@Atreidae - thanks for catch and updating the doc! Welcome to the contributors list! :)

PR #915, opened 2018-10-18
Update using-powershell-to-manage-teams.md

millh0use:
Included information about installing the Teams module from the PowerShell Gallery and changed the manual download link to not point to a specific (outdated) version.


kenwith:
@millh0use - thank you very much for the catch. Pretty embarrassing that the link from the product managers name was right in the article. Appreciate you fixing it. :)

PR #918, opened 2018-10-18
grammar edit

kskarlsson:
Remove the "for" in the following line: "Microsoft partners with selected Session Border Controllers (SBC) vendors to certify that their SBCs for work with Direct Routing"


kenwith:
@kskarlsson - much appreciated!

PR #920, opened 2018-10-18
Need to clarify the NOTE about "CAP account's password will be chang…

cadasi:
…ed to a random password."

I suggested a change, but the (original) text is not clear enough to me in this point " If you have synced an on premises Active Directory to AAD make sure to take note of the credentials you are using that will be changed by CAP provisioning."  Simply because during the provision process, there's no option to provide a password to CAP Service to use. Not sure if CAP should give a password so the SFBO admin can update that onprem. AAD Sync service offers the opportunity to Sync password between Onprem and Online and vice-versa automatically, but not all AAD Sync versions will permit that (older ones will not). It's really confusing. Also, there is a known problem, where if customer is using a third party idp or ADFS for authentication, CAP provision will fail. That's the change I made to the note, but we need to check on the other possible limitations I commented an updated the note to an even better version.


kenwith:
@cadasi - thanks! Much appreciated!

PR #933, opened 2018-10-20
Ross lync tools docs

Ross-Clark:
I fixed the tables in the headers of these docs.


kenwith:
@Ross-Clark - thanks for fix on these tables, much appreciated!

PR #935, opened 2018-10-23
Change link for dial-in and dial-out rates

mrik23:
Changes the link to the correct on to check rates for pay-per-minute Audio Conferencing plan.


kenwith:
@mrik23 - thanks for updating to the Teams rates page. And thanks for making the docs better, much appreciated!

PR #981, opened 2018-10-28
chore: Remove en-us from azure.microsoft.com links

nschonni:


kenwith:
@nschonni - thanks for this fix! Now these will render to the language of the person viewing. Much appreciated!

PR #1001, opened 2018-10-31
Update SharePoint-OneDrive-interact.md

cowitkow:
Website-type tab uses iframes for the web client. SharePoint Online will block this by default and the page will not load. The desktop client may load as it uses webviews. Authentication using the website-type tab is also separate from client authentication, requiring a user to re-auth. In addition, some features will not work in a website-type tab (ex. pop-ups) because of the framed session.

Changed reference from Website-type tab to SharePoint and Document Library tabs.


kenwith:
@cowitkow - confirmed I see the same behavior and confirmed your new procedure works. Thanks for update!

PR #1031, opened 2018-11-07
Remove Early Adoptors / Preview notice

benhopperau:
Removed notice for Early Adopters/ Preview as Telstra Calling for Office 365 has now gone into GA


kenwith:
@benhopperau - thanks for update!

@LolaJacobsen and @CarolynRowe - this looks like removing note that Australia is in preview.. OK to merge?

kenwith:
@benhopperau - thanks for update! Confirmed on link it is GA.

PR #1060, opened 2018-11-13
Update direct-routing-user-accounts-in-a-hybrid-environment.md

carolynblanding:
Updated supportability statement for the hybrid environment. We had redundant information. We don't need to create users in on premise pool and then move. What we must do is enable the users with on premise management tools with hosting provider of online so that on premise is aware of the sip enabled user and their service location.


kenwith:
@CarolynRowe - OK to merge?

PR #1087, opened 2018-11-16
typo: Double word "Server"

nschonni:
BTW, it looks like there is a similar file without the "_1" https://github.com/MicrosoftDocs/OfficeDocs-SkypeForBusiness/blob/live/Lync/LyncServer/configure-federation-routes-and-media-traffic.md


kenwith:
@nschonni - if I recall the _1 was a way to differentiate the Lync Server content from the new 'evergreen' content. I would have to go back and double check the difference but I remember the _1 was a differentiator since we combined a lot of content when we moved from TechNet to GitHub/docs.microsoft.com and some of it wasn't able to be combined so it has a _1. If my memory serves me.

PR #1098, opened 2018-11-19
Sample Script to Automate Firewall Rule Creation

carolynblanding:
We've been getting valid feedback from administrators that they do not want to ask end users to cancel security dialog for Firewall rule with explanation that it will work OK. This is both confusing to end users as well as bad practice to instruct users to ignore security dialogs. Administrators are asking us to please provide a mechanism to allow them to create appropriate firewall rules that will avoid the dialog. 

We have received the following sample script from engineering that works great and avoids the prompt. This script can be run as a GPO computer startup script. Can we add this sample script to this documentation so that administrators know how to control this end user experience.

Sample Script

$users = Get-Childitem c:\users
foreach ($user in $users) 
{
    $Path = "c:\users\" + $user.Name + "\appdata\local\Microsoft\Teams\Current\Teams.exe"
    if (Test-Path $Path) 
	{
	        $name = "teams.exe " + $user.Name
	        if (!(Get-NetFirewallRule -DisplayName $name))
		{
	            New-NetFirewallRule -DisplayName “teams.exe” -Direction Inbound -Profile Domain -Program $Path -Action Allow
		}
	}
}

### DIFF
--- a/Teams/get-clients.md
+++ b/Teams/get-clients.md
@@ -142,3 +142,22 @@ Notification settings
 There are currently no options available for IT administrators to configure client-side notification settings. All notification options are set by the user. The figure below outlines the default client settings.
 
 ![Screenshot of Notifications settings.](media/Get_clients_for_Microsoft_Teams_image6.png)
+
+Sample PowerShell Script
+----------------------------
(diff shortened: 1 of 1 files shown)


kenwith:
@LolaJacobsen - OK to merge?

kenwith:
@carolynblanding - ping on this one... :)

LolaJacobsen:
I'm following up in email with Carolyn. I've asked @jambirk to handle the changes.

PR #1102, opened 2018-11-20
Misspelling

dmadelung:
rganizations to organizations


kenwith:
@dmadelung - thanks for the catch and fix, much appreciated!

PR #1183, opened 2018-12-07
Update let-your-teams-users-communicate-with-other-people.md

carolynblanding:
Making it clearer to customers how to limit communications to other organizations either with allow all but blocked, or limit communications to specific organizations. 

Proposed changed based on feedback from support personnel.


LolaJacobsen:
Ping'd @aspiziri-ms offline for review.

LolaJacobsen:
Ping'd @Karvell offline - approved.

PR #1189, opened 2018-12-10
Update direct-routing-configure.md

AlexanderHolmeset:
Was using set-csuser thats for on-premises skype, should have used set-csonlineuser. Also forgot to write -Indentity.


kenwith:
@AlexanderHolmeset - thanks for catch on that!

PR #1210, opened 2018-12-12
Produce Skype Meeting Broadcast

naminakao:
Based on the following public content, SfB 2015 client is also able to produce Skype Meeting Broadcast if the client build is 15.0.4747 or later. And also SfB 2016 client has the build requirement,  C2R 16.0.4227 or later.
https://support.office.com/en-us/article/c472c76b-21f1-4e4b-ab58-329a6c33757d


kenwith:
@PhillipGarding @jambirk - OK to merge this?

PhillipGarding:
That change looks correct.

Phil.

PR #1212, opened 2018-12-12
Sync delegates btwn SfB & OL

naminakao:
Based on the following content, SfB 2015 client seems also support synchronize delegate between SfB and Outlook.
https://docs.microsoft.com/en-us/skypeforbusiness/set-up-skype-for-business-online/set-up-and-troubleshoot-skype-for-business-online-delegation


kenwith:
@PhillipGarding @jambirk - OK to merge this?

kenwith:
@PhillipGarding and @jambirk - ping on this one too. It has been open for 29 days.

PhillipGarding:
I just don't know on this one.  It was not checked in the Server 2015 version of this topic, and that table was populated by people more familiar with SfB 2015 client behavior.

jambirk:
@kenwith Maybe we need to revert this if @PhillipGarding can't confirm.

kenwith:
I can revert, I read through the document listed above (in the comment for the PR) and it looks like what @naminakao added is correct according to the document, or am I missing something?

jambirk:
@PhillipGarding has final say on technical accuracy. It's his call.

PhillipGarding:
The linked documentation doesn’t explicitly say that synchronization will happen, but there is an attribute (EnableExchangeDelegateSync) that certainly seems like this is what it is talking about.  The synchronization is done in the background, and it means that if you enable a delegate in Outlook, you don’t have to go into SfB options and enable it as a delegate.  The person will eventually be listed automatically in the SfB delegate list (although you will have to choose to route calls to delegates, etc.)  This article tells you to go into SfB options and manually set this, which you wouldn’t need to do if sync was happening.  Chances are that sync was implemented in 2015, so we could list it.

Phil.

PR #1213, opened 2018-12-12
Give/take control of screen sharing

naminakao:
SfB 2015 is able to give/take control of screen sharing. The following content does not mention SfB client version but it is created at Feb.04, 2015. Assume SfB 2015 supports to give/take control of screen sharing.
 https://support.office.com/en-us/article/2d436dc9-d092-4ef1-83f1-dd9f7a7cd3fc


kenwith:
@naminakao - thanks for making the content better!

@PhillipGarding  @jambirk - OK to merge this?

kenwith:
@jambirk

PhillipGarding:
I don't know if this was supported in 2015 client or not.  I would assume so, but I don't know.  The SErver 2015 version of the client comparison tables did NOT have this checked, and that table was reviewed by people more knowledge about 2015 client behavior.

Issue #161, opened 2018-05-01
Can see Heartbeats but no SRS-App Events


danielmarker:
After adding the "Skype Room System" into Windows Logs and installing the agent - I am unable to see any events apart from Heartbeats, any troubleshooting steps available for this which could be added to this document?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: ef7ceb43-d37f-8918-855e-ac8375875e6e
* Version Independent ID: 1b48ad59-2737-7ac4-8ad5-80904c0e3ccc
* Content: [Deploy Skype Room Systems v2 management with OMS](https://docs.microsoft.com/en-us/skypeforbusiness/deploy/deploy-clients/with-oms)
* Content Source: [Skype/SfbServer/deploy/deploy-clients/with-oms.md](https://github.com/MicrosoftDocs/OfficeDocs-SkypeForBusiness/blob/live/Skype/SfbServer/deploy/deploy-clients/with-oms.md)
* Service: **unspecified**
* GitHub Login: @jambirk
* Microsoft Alias: **jambirk**

jambirk:
Turgay will have to be the one to answer this, but thanks for bringing it to our attention.

Turgayo:
Hey Daniel,
Can you confirm that the Heartbeat events you are getting are from the Skype Room System event log, not the generic heartbeat messages generated by the OMS agent itself.

To do that, can you please check if the query below returns any data?
Event | where Source == "SRS-App"

PR #1215, opened 2018-12-12
Update Known-issues.md

MarcoCarampanta:
Hi,
I purpose a change to the line #161 as our internal information mentions both Strings actually. This is also confirmed on our internal info by the SfB/Teams EEE - Roman Maddox, and Teams Senior SEE - Carolyn Blanding.

Additionally, on the powershell cmdlet that was added before, it would in fact replace the ENTIRE EWSAllowList, with just *skypespaces* this can be very very critical to the customer's environment as we won't have a way to recover what he had before!

The cmdlet I added: Set-organizationconfig -EwsAllowList @{Add="*MicrosoftNinja*","*SkypeSpaces*"} ;
Is effectively just **adding** those 2 strings, instead of **replacing** in its entirety.

Note: this type of issue is arising more and more, and we're getting multiple tickets which lead to confusion when we discuss it with the customers and direct them here, where only *skypespaces* is mentioned (which won't fix the meeting access issue).

### DIFF
--- a/Teams/Known-issues.md
+++ b/Teams/Known-issues.md
@@ -158,7 +158,7 @@ This article lists the known issues for Microsoft Teams, by feature area.
 
 |**Issue title**|**Behavior / Symptom**|**Known workaround**|**Discovery date**|
 |:-----|:-----|:-----|:-----|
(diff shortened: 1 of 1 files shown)


LolaJacobsen:
@marclauinger - please review. Let me know if we can merge this change.

marclauinger:
@LolaJacobsen and @MarcoCarampanta, I took a look and this is fine to be merged.

PR #1232, opened 2018-12-13
Update example domain

erwinbierens:
changed example domain to contoso, in the examples contoso is also used. 
abc.xyz is confuzing for customers.


kenwith:
@erwinbierens - thanks for making this clearer! much appreciated.

PR #1261, opened 2018-12-18
Removed references to 'watcher3'

greiginsydney:
Content now aligns with [Set-CsWatcherNodeConfiguration](https://docs.microsoft.com/en-us/powershell/module/skype/set-cswatchernodeconfiguration?view=skype-ps)
"You must always have at least two test users per watcher node." 
In operation, Watcher Nodes only ever use a maximum of 2 users per test (even conferences).


AndreaBarr:
@jambirk @CarolynRowe @MicrosoftHeidi Can I merge this?

PR #1263, opened 2018-12-18
Remove references to PartitionByOU and msRTCSIP-GroupingID

sorindut:
Per ICM 93681635 and ICM 27777378, the address book partitioning no longer works properly - mainly because the desktop client bypasses it through PCX search. Having the reference to address book partitioning creates wrong expectations.


AndreaBarr:
@jambirk @CarolynRowe @MicrosoftHeidi Can I merge this?

PR #1290, opened 2018-12-20
Update set-auto-attendant-languages-for-audio-conferencing-in-teams.md

ruizoscar:
Updating the document to correct a couple of sentences that were incorrectly pointed out which phone number can have their languages changed.


AndreaBarr:
@LolaJacobsen @chuckedmonson @tonysmit Can I merge this?

PR #1291, opened 2018-12-21
removed info where onprem users are able to create Skype Brodcast mee…

savusca:
…tings

Creating a Skype Broadcast meeting requires user to be licensed with at least Skype Online Plan 2 license. Granting such license to an on-prem hosted account can lead to federation/unexcected behavior. Removed information where is stated that onprem accounts can create Skype Broadcast meetings.


AndreaBarr:
@LolaJacobsen @tonysmit Can I merge this?

PR #1330, opened 2018-12-28
Removed reference to Solutions Catalog.

v-calem:


jambirk:
@v-calem did you mean to make that mention of monitors another list item, or format it as italic? please view again in preview before I publish

PR #1342, opened 2019-01-02
Add details for hosting vs streaming licenses required

micchan:
Add a note that attendees streaming a skype meeting broadcast do not require any user licenses.  Only users that would like to host a broadcast event requires a valid O365 license


AndreaBarr:
@LolaJacobsen @tonysmit Can I merge this?

kenwith:
@AndreaBarr - I looked in the article and @micchan was listed as the ms.reviewer so I went ahead and merged.

PR #1388, opened 2019-01-12
Unassign button is greyed-out

CarlosFarrica:
Troubleshooting steps


AndreaBarr:
@LolaJacobsen @tonysmit @chuckedmonson Can I merge this?

kenwith:
@CarlosFarrica - thanks for this support tip! It seems very valuable for people to see this so they can solve it before calling support. Let's loop @oscarr and @tonysmit in to confirm and then I will get this merged and live.

PR #1454, opened 2019-01-22
Replaced "gcm" in example with Get-Command

greiginsydney:
I'm told aliases are naughty.


kenwith:
@greiginsydney using the full Get-Command makes it clearer too, thanks for fix!

PR #1474, opened 2019-01-24
Update set-up-audio-conferencing.md

AlexanderHolmeset:
You cant choose numbers to edit under Microsoft Bridge Settings, its under Microsoft Bridge menu.


kenwith:
@AlexanderHolmeset thanks for catch and fix, much appreciated!

PR #1497, opened 2019-01-29
Fixed md bugs

greiginsydney:
Reformatted for clarity, splitting the supported and unsupported scenarios into separate tables. (Github's MD doesn't provide an equivalent to "colspan" - for that your only option is to rebuild the table as HTML.)


kenwith:
@greiginsydney - thanks for formatting fix, much appreciated!

PR #1498, opened 2019-01-29
Update direct-routing-user-accounts-in-a-hybrid-environment.md

shanehoey:
Fixed Table Layout,  was showing as text only on docs due to missing | and additional line breaks


kenwith:
@shanehoey - thanks for formatting fix, much appreciated!

PR #1505, opened 2019-01-30
Update direct-routing-trunk-failover-on-outbound-call.md

shanehoey:
ReinviteResponceCode is not a valid parameter on current sfbo module, -FailoverResponseCodes should be used instead, also line 47 refered to it as ReinviteResponceCode  but line 51 refered to it as ReinviteResponceCode , so have fixed that as well


scanum:
@CarolynRowe can this be merged?

PR #1521, opened 2019-01-31
Updated QoS GPO step to add the Application Name

greiginsydney:
This change now aligns the example with the preface text & the registry extract in Table 4, which both reference the Teams client specifically.
Also added the overlooked "repeat for video & app sharing" steps.


AndreaBarr:
@LolaJacobsen @tonysmit @chuckedmonson Can I merge this?

PR #1535, opened 2019-02-04
Deleted the "nested" reference in the holiday handling

greiginsydney:
(This looks to be a copy/paste oversight from the earlier references to same, in the business hours and non-business hours menus).


jambirk:
This file is currently under rework for other reasons, we can't accept this change at this time.

PR #1549, opened 2019-02-05

carolynblanding:
Updated table:
1. Change Retention support to Yes with supplemental information item 3 for on dedicated and on premise exchange.
2. Added supplemental information note 4 for voicemail to provide more context. Voicemail is supported in Teams regardless of mailbox location, but voicemail messages will not be available in Teams for on premise mailbox. 
3. Removed (hybrid configuration) for requirement for Oauth for Exchange on-premise. Oauth MUST be configured so that exchange on premise allow Outh authenticated connection to EWS from Teams Middle Tier Service. HOWEVER, Houston we have a problem, the Exchange article we are referencing has the WRONG authserver configuration. Teams service will present a token from Azure AD (evoSTS) and that needs to be configured in Exchange on premise for authentication to be successful. Without this, connectivity to mailboxes to access user data will fail. I put in a github request to have that article updated but not sure if that will get processed. If you could put in a word to that team, that would be awesome.


carolynblanding:
I need to make a correction to the above. What we need to emphatically state is that Exchange Hybrid is a requirement. Helped an engineer on a case and found that we need the exchange web service URL's in the Service Principal names in online as well as the on premise configured with the Oauth Server. The Exchange Hybrid Wizard will handle this, but we also need to state that we require Exchange 2013 latest CU-1 and customer should re-run the Hybrid Wizard if they had an older Exchange 2013 deployment. 

So on the 3rd update I made, we actually need to change the whole reference to Oauth and instead just state what I said above. We can skip the reference out to the outdated Exchange Oauth article.

Issue #990, opened 2018-10-29
SfB Server 2019 Environmental requirements for Skype for Business Server link incorrect 


Markusott7:
Environmental requirements for Skype for Business Server link incorrectly points to 2015 requrements instead of 2019 requirements:

At: Create a file share in Skype for Business Server
https://docs.microsoft.com/en-us/skypeforbusiness/deploy/install/create-a-file-share

Incorrect: Environmental requirements for Skype for Business Server
https://docs.microsoft.com/en-us/skypeforbusiness/plan-your-deployment/requirements-for-your-environment/environmental-requirements

Correct: System requirements for Skype for Business Server
https://docs.microsoft.com/en-us/skypeforbusiness/plan/system-requirements

AndreaBarr:
@Markusott7 Thank you for this feedback

I will get this information about the wrong links to the Skype for Business team for investigation.

The team will get this prioritized into a sprint plan. Thank you for reporting and making the docs better. Much appreciated.

I made a note to request the team to update this when the work is complete.

AndreaBarr:
@kellipav could you please make the changes listed above?

MicrosoftHeidi:
Environmental requirements is a 2015 article. System Requirements is the 2019 article. There should be nothing in the 2015 article that references 2019.

jambirk:
@MicrosoftHeidi , this is a deploy article that is shared by both versions. There needs to be a link to both articles.

@Markusott7 the links provided correctly point to different articles: one for server 2015 requirements and one for server 2019 requirements. This is because the article is shared by both versions, and the process is only different in cases like this, when separate links are needed to get to version-dependant requirements. Docs.microsoft.com lets us do certain things that weren't possible on Technet.microsoft.com, and this is one of them.

PR #1563, opened 2019-02-07
Fixing System Reqs Link

get-itips:
Fixing System Reqs Link according to https://github.com/MicrosoftDocs/OfficeDocs-SkypeForBusiness/issues/990 @yogkumgit


jambirk:
@MicrosoftHeidi should approve this.
all the links in the changed section still look malformed to me.

MicrosoftHeidi:
This should not be merged, those articles aren't in the 2019 branch and so are incorrect.

SerdarSoysal:
@jambirk there seems to be a TOC issue with this topic in the 2019 TOC. When you click on the "create a file share" link in the 2019 TOC, you fall back into the 2015 TOC, and in that context the links in the article are correct as @MicrosoftHeidi says. Can you please investigate what needs to be done to fix the 2019 TOC?

jambirk:
Agree with Heidi this should not be merged. This is an evergreening issue/misunderstanding: the article in question is used by both versions, and depending on which version of SfB the reader needss to install they need to go to different requirements topics, which are not shared. having a look at the TOC next.

get-itips:
Thanks @jambirk for clarification.

PR #1610, opened 2019-02-11
Update get-clients.md

baardhermansen:
Updated the PowerShell script. I know it says it's a sample, but why not present a more robust sample? :)
Added synopsis and a bit more robustness. Specifies requires PS version 3, excludes local admins and Public users, looks for a firewall rule containing the path to Teams.exe for the users instead of a firewall rule name. Specifies TCP and UDP port instead of allowing all ports.

### DIFF
--- a/Teams/get-clients.md
+++ b/Teams/get-clients.md
@@ -146,29 +146,36 @@ Sample PowerShell Script
 This sample script, which needs to run on client computers in the context of an elevated administrator account, will create a new inbound firewall rule for each user folder found in c:\users. When Teams finds this rule, it will prevent the Teams application from prompting users to create firewall rules when the users make their first call from Teams. 
 
 ```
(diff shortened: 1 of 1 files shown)


scanum:
@LolaJacobsen can this be merged?

jambirk:
@LolaJacobsen is OOF. @scanum  I don't see a problem with publishing it, but I really think @tonysmit should be the one to approve, under the circumstances.

scanum:
@tonysmit since everything looks correct, can be this merged?

tonysmit:
@LolaJacobsen - You will need to confirm this one. Thanks.

LolaJacobsen:
I'm conferring with the PM and EM for this feature - awaiting their reply.

LolaJacobsen:
Per EM: "I've reviewed the script changes and it looks good."

Issue #1575, opened 2019-02-08
Org-wide teams


MikeParker365:
Maximum number of users in teams is now 5,000

---
#### Document details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 06059d07-6687-2924-7d0a-3c03220a1650
* Version Independent ID: bb293f92-9394-b156-2e08-0baead5c3bbe
* Content: [Overview of teams and channels in Microsoft Teams](https://docs.microsoft.com/en-gb/MicrosoftTeams/teams-channels-overview)
* Content Source: [Teams/teams-channels-overview.md](https://github.com/MicrosoftDocs/OfficeDocs-SkypeForBusiness/blob/live/Teams/teams-channels-overview.md)
* Service: **msteams**
* GitHub Login: @LolaJacobsen
* Microsoft Alias: **lolaj**

PR #1652, opened 2019-02-19
Update teams-channels-overview.md

RussC19:
Updated 2500 members to 5000, as per https://github.com/MicrosoftDocs/OfficeDocs-SkypeForBusiness/issues/1575

One point to note, MikeParker365, and to make it easier for the editor is to to keep with the terminology used in the document when reporting errors or amendments. In this instance, it's straightforward as though [unlimited] users could be invited there's only a finite number of members (if that makes sense). Thanks, Mike.

### DIFF
--- a/Teams/teams-channels-overview.md
+++ b/Teams/teams-channels-overview.md
@@ -2,6 +2,7 @@
 title: Overview of teams and channels in Microsoft Teams
 author: LolaJacobsen
 ms.author: lolaj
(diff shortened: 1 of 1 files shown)


scanum:
@SerdarSoysal @tonysmit Changes look correct, can this be closed?

tonysmit:
@LolaJacobsen - I think this is correct, but you will need to confirm. Thanks.

PR #1691, opened 2019-02-25
Remove references to old discontinued Call Analytics preview portal

stephenjust:


kenwith:
Confirmed, and merged. @stephenjust - thanks, much appreciated!

PR #1693, opened 2019-02-25
Update to include known Conditional Access issue

lokimeyburg:
There is a known issue where users who have Conditional Access rules may not be able to view a website tab in the Teams desktop client. This is because the Teams website tab is not recognized as a valid browser when using Conditional Access where browser restrictions are common. The only workaround is to use the browser version of Teams instead of the desktop version.

### DIFF
--- a/Teams/Known-issues.md
+++ b/Teams/Known-issues.md
@@ -3,7 +3,7 @@ title: Known issues for Microsoft Teams
 author: LolaJacobsen
 ms.author: lolaj
 manager: serdars
-ms.date: 2/19/2019
+ms.date: 2/25/2019
 ms.topic: article
 ms.service: msteams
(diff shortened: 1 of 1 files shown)


kenwith:
I believe the was another PR that addressed this. @cowitkow submitted it and I just merged it. If this should also be merged let me know and please go ahead and re-open. Sorry for delay on these PR's that led to two with the same issue. Thanks for making the docs better, much appreciated!

cowitkow:
@kenwith, @lokimeyburg 's is a bit different. I re-added a wider item against Website-type tab's and their usability (iframes/webviews) impacting customer experience. Loki's is specific to conditional access

kenwith:
@lokimeyburg @cowitkow - ok got you, will merge this one too.. let me know if any updates needed.

PR #1706, opened 2019-02-27
update this doc for app setting migration

ritika2502:
@LanaChin we should update this doc and link to the app migration, app setup and app overview docs


kenwith:
@ritika2502 - this is actually a Pull Request (PR) so if I merged it the note would be in there. Probably better to send a chat or email to @LanaChin . :)

PR #1717, opened 2019-02-28
Fixing broken link

TokyoScarab:
fixed broken link(404) to point to similar helpful page


kenwith:
Please make sure /en-us/ is not in any links as that is a language specifier. Without a language specifier in URL the page will open in whatever language the users web browser is set to use.

Let's confirm with @CarolynRowe if the link is correct too.

jambirk:
There have been a few cases lately of links that HAVE to go to en-us or they 404, usually  links to the product catalog.

kenwith:
@jambirk - seems to work fine for me: https://docs.microsoft.com/skypeforbusiness/skype-for-business-hybrid-solutions/plan-your-phone-system-cloud-pbx-solution/plan-phone-system-with-on-premises-pstn-connectivity    Are you getting a 404?

jambirk:
No, that one works okay for me. I think it's just links away from docs but still within microsoft we may need to worry about.

PR #1746, opened 2019-03-07

carolynblanding:
We need to update this article to match the updated Exchange Ouath configuration steps here: https://docs.microsoft.com/en-us/exchange/configure-oauth-authentication-between-exchange-and-exchange-online-organizations-exchange-2013-help

I would suggest that instead of duplicating the steps that we reference that article for configuring the Oauth for Exchange and break out in this article just those steps that are specific to the SfB -> Exchange Oauth.


scanum:
@jambirk Hi James, could you please check and merge? Changes look correct to me. Thanks!

jambirk:
@scanum  this isn't anywhere near ready to publish. 1. We shouldn't publish raw urls like this, it need to be a properly formatted link. 2. inserting a paraphrase of the comment is not a solution. 3. inserting anything above the H1 title like that just skews the article formatting.

jambirk:
@scanum this is not publishable as a solution. closing this PR

PR #1798, opened 2019-03-15
Update Known-issues.md

cowitkow:
SPO vanity domain issue resolved February 2019. https://icm.ad.msft.net/imp/v3/incidents/details/104609094/home

### DIFF
--- a/Teams/Known-issues.md
+++ b/Teams/Known-issues.md
@@ -247,10 +247,6 @@ This article lists the known issues for Microsoft Teams, by feature area.
 
 |**Issue title**|**Behavior / Symptom**|**Known workaround**|**Discovery date**|
 |:-----|:-----|:-----|:-----|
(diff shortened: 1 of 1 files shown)


LolaJacobsen:
@marclauinger needs to approve any changes to this article. Please do NOT merge this PR without his approval.

marclauinger:
The ICM attached does not match the known issue.  The ICM is about adding a sharepoint tab and the Save button not getting activated, nothing about when going to the new tab it showing a blank page.  Has someone verified that adding an SPO page to a tab actually renders correctly now?

cowitkow:
That should be relevant ICM that was holding up the actual deployment of the fix. This is also documented against https://domoreexp.visualstudio.com/MSTeams/_workitems/edit/287249 and https://domoreexp.visualstudio.com/MSTeams/_workitems/edit/243161. We can also defer to rajgopal@microsoft.com for this as needed.

marclauinger:
Then it should be good.  I couldn't find the bugs and in the ICM I didn't see any links.  Approving.

PR #1841, opened 2019-03-20
Fix code formatting

robdy:


scanum:
@bjwhalen Hi! Could you please check this PR and merge if correct? Thanks!

jambirk:
@scanum 
I know it's not really my call, but the change here is solely to the formatting and not the command itself. I just don't see why this needs to be a code block that readers can copy/paste. Having a single tick mark enclose the output text is correct formatting according to my understanding.

jambirk:
the fix as implemented was incorrect. Should work now.

PR #1857, opened 2019-03-21
Update set-up-guests.md

expiscornovus:
Removed the Tbot setting from the example, Tbot has been removed:
https://tomtalks.blog/2018/10/microsoft-teams-are-killing-t-bot-new-help-experience-coming/


LolaJacobsen:
@islubin and @somakbhattacharyya - could you review this PR please? If we are removing the AllowTBotProactiveMessaging cmdlet from this article, should it be removed from the PowerShell cmdlet documentation too?

islubin:
Yes, this PR is fine.  I'll update the clientconfiguration powershell doc as well.

islubin:
I checked and realized that I'd already highlighted that value as "Deprecated, do not use" so we should be good.

PR #1860, opened 2019-03-21
Author is GitHub alias. Updated.

kenwith:


kenwith:
@scanum - FYI and for review.. I think Lola GitHub is different than Microsoft alias.

Issue #1848, opened 2019-03-21
The link to this PowerShell reference fails from the main Teams page


ChrisLGardner:
On the main Teams overview on Docs (https://docs.microsoft.com/en-gb/MicrosoftTeams/Microsoft-Teams#pivot=home&panel=home-all) the PowerShell link is a 404 (https://docs.microsoft.com/en-gb/powershell/module/teams/?view=teams-ps) and should almost certainly go to this page instead. The suggestions even point to this page.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 26362d70-cfce-2eea-077b-4aaa50f92e8b
* Version Independent ID: f52dcd4d-4f58-05c4-11b0-f1e3c95f4d89
* Content: [Teams PowerShell Overview](https://docs.microsoft.com/en-us/microsoftteams/teams-powershell-overview#feedback)
* Content Source: [Teams/teams-powershell-overview.md](https://github.com/MicrosoftDocs/OfficeDocs-SkypeForBusiness/blob/live/Teams/teams-powershell-overview.md)
* Service: **msteams**
* GitHub Login: @LolaJacobsen
* Microsoft Alias: **lolaj**

scanum:
@officedocsbot assign @scanum

sowjanya365:
Hi @ChrisLGardner 

Thanks for reaching out!

We see that the issue is resolved now. Link is working fine and taking us to the correct URL. Please check and let us know if there is any issue. Thank you.

ChrisLGardner:
That link now works correctly. 

Thanks

PR #1877, opened 2019-03-24
Update Microsoft-Teams.md index due to 404 on Powershell cmdlet reference

shanehoey:
Updated page due to 404 on Powershell Cmdlet reference, 

I think it should be going to the  Teams PowerShell Overview, which includes information on both teams and sfbo modules

### DIFF
--- a/Teams/Microsoft-Teams.md
+++ b/Teams/Microsoft-Teams.md
@@ -178,7 +178,7 @@ f1keywords:
                             </a>
                         </li> 
                         <li>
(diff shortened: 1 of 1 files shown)


shanehoey:
Just realised this is also logged as an issue #1848

kenwith:
Yes, was down the end of last week, then there was an issue opened in PowerShell repo.. then it was fixed for all but Exchange. Now all these TOC references are down again. :(  I opened a case with the team that owns this and they are working on it now.

kenwith:
Should be fixed now.

PR #1878, opened 2019-03-25
Update Tutorial-Calling-in-Teams.yml

shanehoey:
Markdown Fixes only

1) fixes the Markdown in the Envision section, for example  ###Connecting to the PSTN  should be ### Connecting to the PSTN
2) fixes the Markdown in the Onboard Section for example ###1. Enable Exchange on-premises for Azure Voicemail should be ### 1. Enable Exchange on-premises for Azure Voicemail
3) fixes the ordered list that is displaying steps listed as 0 thru to 8 instead of 1-9

### DIFF
--- a/Teams/Tutorial-Calling-in-Teams.yml
+++ b/Teams/Tutorial-Calling-in-Teams.yml
@@ -221,7 +221,7 @@ items:
 
     -   Optional: Dial plans
 
-    ###Connecting to the PSTN
+    ### Connecting to the PSTN
 
     In order for users to place and receive calls to the PSTN by using Teams, you
     need to have installed Phone System and you need to connect phone system to the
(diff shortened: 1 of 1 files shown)


kenwith:
@shanehoey - thanks for the PR and surprised to see it so soon after you headed back home! I think I would still be asleep if I had the flight you did. :)

@LolaJacobsen - this article looks like it was written by Thomas so  @scanum will probably be looking for new author that is taking this over for Thomas. Would that be you?

LolaJacobsen:
@shanehoey thanks for your contribution, and it was so nice to meet you last week!

@kenwith I'll work with Thomas to figure out a new owner and update his articles accordingly.

shanehoey:
@Kenwith and @LolaJacobsen  Great to meat you both as well, expect to see lots more PR's from me :)

PR #1898, opened 2019-03-27
Removed an unnecessary reference to PSTN IP

robertpp:
Removed an unnecessary reference to PSTN IP in the multi-homed note, which made it too specific.


scanum:
@jambirk Hi James, could you please check this PR? If you agree with the change could you merge? Thanks!

jambirk:
I don't agree. The sentence as written ties the note to the preceding paragraph, and should be kept as is.

jambirk:
@robertpp updated with a different change to resolve issue. Do you agree with this clarification?

PR #1916, opened 2019-03-29

shanehoey:
The current format of the script to log into powershell is not consistent with other parts of Docs.ms,  in this page the script block is separate for each cmdlet, and inside a note block. 

In this proposed changed I have separated the note and the powershell to log onto sfbo.  In the current format its hard to tell that there is actual powershell  in the note.


scanum:
@tonysmit Hi Tony, changes look correct. Could you please confirm and merge? Thanks!

Issue #1865, opened 2019-03-22
Missing critical step


oldkingcolemsscna:
After step 10, this article should refer the reader to install Skype For Business Server 2019 before proceeding with the "Verify pilot pool..." step.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 1ec4d122-ff87-2c4b-6caa-299795190dc5
* Version Independent ID: de2481bd-8b21-b5d7-3898-1a4b40d0bfd2
* Content: [Deploy pilot pool](https://docs.microsoft.com/en-us/skypeforbusiness/migration/deploy-pilot-pool#feedback)
* Content Source: [Skype/SfBServer2019/migration/deploy-pilot-pool.md](https://github.com/MicrosoftDocs/OfficeDocs-SkypeForBusiness/blob/live/Skype/SfBServer2019/migration/deploy-pilot-pool.md)
* Product: **skype-for-business-itpro**
* GitHub Login: @kenwith
* Microsoft Alias: **kenwith**

scanum:
@officedocsbot assign @scanum

get-itips:
Hello @oldkingcolemsscna 
Thanks for your feedback, we really appreciate it.
I created a Pull Request to improve the article adding the missing step.
Thank you

scanum:
@oldkingcolemsscna Thank you for submitting this feedback.

From our understanding the issue has been resolved. If you feel it hasn't been resolved please re-open this issue.

@officedocsbot close

PR #1923, opened 2019-03-31
Added missing step to install SfB binaries

get-itips:
Added missing step to install SfB binaries after publishing topology and before verifying coexistence.

More info: https://github.com/MicrosoftDocs/OfficeDocs-SkypeForBusiness/issues/1865


scanum:
@kenwith Hi Ken! Changes look correct. Could you please confirm and merge? Thanks!

PR #1964, opened 2019-04-04
adding Microsoft infront of Teams

siunies:


LolaJacobsen:
Don't need to prepend Teams with Microsoft except on first use. Closing without merging.

Issue #1840, opened 2019-03-20
First party vs third party


tmaheshwar:
First party templates are these going to appear via General Tab, which are 3rd party Templates

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: eb17d752-d62d-1d41-fa98-5eafe2c057d9
* Version Independent ID: 3ced31b7-2ac5-e140-b848-f97eee63a52f
* Content: [Get started with Teams templates for Healthcare organizations](https://docs.microsoft.com/en-us/microsoftteams/healthcare/healthcare-templates#feedback)
* Content Source: [Teams/healthcare/healthcare-templates.md](https://github.com/MicrosoftDocs/OfficeDocs-SkypeForBusiness/blob/live/Teams/healthcare/healthcare-templates.md)
* Service: **msteams**
* GitHub Login: @jambirk
* Microsoft Alias: **jambirk**

scanum:
@officedocsbot assign @scanum

scanum:
@tmaheshwar  Thank you for submitting this feedback.

From our understanding the issue has been resolved. If you feel it hasn't been resolved please re-open this issue.

@officedocsbot close

BakerTeams:
IS there a step by step guide for the templates. I have been trying to get them to import using powershell and cannot get it.

PR #1995, opened 2019-04-08
Update healthcare-templates.md

sukihanabi:
#1840 
Does the user mean "where the channels in the template will appear"?
If that is his question, the answer is: "The channels in the template will automatically be created under the General Tab."

Action Taken: Added Note

Did the ff:
Test the Template in standard teams "https://graph.microsoft.com/beta/teamsTemplates('standard')"
these created channels under the general tab, see screenshot https://snag.gy/G93Ue5.jpg

Reference:
https://www.sharepointnutsandbolts.com/2019/03/create-microsoft-teams-from-template.html


BakerTeams:
I guess I am missing something. Is there a step by step guide to get these templates into teams?

PR #2079, opened 2019-04-19
updated aad image

SebastianSchuetze:


LolaJacobsen:
@SharS - this is an image you worked on recently. Would you review this contributor's proposed changes and advise on whether to accept or not?

scanum:
@LolaJacobsen

kenwith:
@LolaJacobsen - just FYI - I asked @scanum to tag you so this shows up in your query of @ mentions for follow up. Just FYI in case you wonder why you were tagged there. :)

SharS:
Thank you for the new image. It will need to go through our art process before it can be added to the article. That process can take up to 2 weeks.

Issue #1707, opened 2019-02-28
Dead link - pricing


jason-ra:
Link to pricing page is 404: https://products.office.com/en-AU/skype-for-business/audio-conferencing#Requirements

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 5c6e2b61-656c-a4f8-94a7-ec3c14e4d65e
* Version Independent ID: 77932de0-0312-95c5-f356-89ece9b3aabd
* Content: [Phone numbers for Audio Conferencing in Microsoft Teams](https://docs.microsoft.com/en-us/microsoftteams/phone-numbers-for-audio-conferencing-in-teams)
* Content Source: [Teams/phone-numbers-for-audio-conferencing-in-teams.md](https://github.com/MicrosoftDocs/OfficeDocs-SkypeForBusiness/blob/live/Teams/phone-numbers-for-audio-conferencing-in-teams.md)
* Service: **msteams**
* GitHub Login: @tonysmit
* Microsoft Alias: **tonysmit**

scanum:
@officedocsbot assign @scanum

lindspea:
@jason-ra Thank you for pointing this out.  I have just checked and the link does work.  Please confirm if you're still having an issue with the link.

jason-ra:
Still not working for me. I've just checked and the actual link in the doc is https://products.office.com/skype-for-business/audio-conferencing which returns a "302 Moved Temporarily" that redirects my browser to https://products.office.com/en-AU/skype-for-business/audio-conferencing.

If I replace 'en-au' with 'en-gb' or 'en-us' it does work. So just seems to be an issue with Australian locale 'en-au'

kenwith:
@jason-ra - I wonder if it could be a caching issue? Can you try in an 'in-private' browsing window so that it doesn't look at any cache? I am trying this link: https://products.office.com/en-au/microsoft-teams/online-meeting-solutions#Requirements and it is rendering for me.

There is also a localization team we can check with if it still isn't working for you. Or perhaps something in the redirect was just fixed that I am not aware of. We will get this sorted out though and really appreciate you taking the time to submit the issue to make the docs better!

jason-ra:
[...]

PR #2169, opened 2019-05-03
Update phone-numbers-for-audio-conferencing-in-teams.md

lindspea:
https://github.com/MicrosoftDocs/OfficeDocs-SkypeForBusiness/issues/1707
Updated link - added en-us as per Office Global.


tonysmit:
This isn't correct. Remove en-us from every link! If you don't localization will create a loc bug to be resolved because the target URL will always appear in en-us no matter of the browser language setting.

I fixed the URL with the correct one:  
https://go.microsoft.com/fwlink/?linkid=799762

lindspea:
Thanks Tony. This was a very confusing one.  The loc team advised to add
the en-us to the link so there was much confusion going on.  Thanks for
correcting it and sorting it out.
Regards,
Lindsay
[image: Picture]

PR #2170, opened 2019-05-03
Update manage-resource-accounts.md

waseemhashem:
@jambirk


tonysmit:
@jambirk - There is no such thing as a "Calling Plan service number". It's a "toll or toll-free service number". Please correct this.

CC: @waseemhashem

Issue #1461, opened 2019-01-22
Get Attached Devices - Missing Logitech MeetUp


rogersau:
The get attached devices script seem to miss the Logitech MeetUp Camera. As in device manager this is under the PNPClass of "Camera" rather than "Image"

Suggest updating to; 
gwmi -Class Win32_PnPEntity | where {$_.PNPClass -eq "Image" -or $_.PNPClass -eq "Camera"}

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: f68cba21-b43e-ed44-449c-da752f5a0f53
* Version Independent ID: 5a2a2c5e-ad0e-833d-6ab7-74ed26f63d4a
* Content: [Skype Room Systems v2 maintenance and operations](https://docs.microsoft.com/en-us/skypeforbusiness/manage/skype-room-systems-v2/room-systems-v2-operations)
* Content Source: [Skype/SfbServer/manage/skype-room-systems-v2/room-systems-v2-operations.md](https://github.com/MicrosoftDocs/OfficeDocs-SkypeForBusiness/blob/live/Skype/SfbServer/manage/skype-room-systems-v2/room-systems-v2-operations.md)
* Product: **skype-for-business-itpro**
* GitHub Login: @jambirk
* Microsoft Alias: **jambirk**

scanum:
@officedocsbot unassign @yogkumgit

lindspea:
@rogersau Thank you for pointing this out and helping improve MS Docs. I have updated the wording as per your suggestion.

jambirk:
We've checked with the product group, and this change to the commands is inadvisable. We'll leave this as it's currently written.

PR #2180, opened 2019-05-06
Update room-systems-v2-operations.md

lindspea:
https://github.com/MicrosoftDocs/OfficeDocs-SkypeForBusiness/issues/1461
Changed wording from 'image' to 'camera'

### DIFF
--- a/Teams/room-systems/room-systems-v2-operations.md
+++ b/Teams/room-systems/room-systems-v2-operations.md
@@ -120,7 +120,7 @@ To Get Attached Devices
   
 ```
 invoke-command {Write-Host "VIDEO DEVICES:" 
-gwmi -Class Win32_PnPEntity | where {$_.PNPClass -eq "Image"} | Format-Table Name,Status,Present; Write-Host "AUDIO DEVICES:" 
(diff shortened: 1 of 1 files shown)


scanum:

scanum:
@jambirk changes look correct. Could you please confirm and merge? Thanks!

jambirk:
This is sample code meant to be copy/pasted to an admin console.  Why are we updating this?

Travis-Snoozy:
That is not script logic I maintain. Take the following casual observations with grains of salt.

- #1461 notes **a logical OR**. The change, however **replaces** "Image" with "Camera" -- which will make all the previously-listed image devices no longer be listed, and only the camera devices show. This is **probably** not the behavior you want. 
- I don't know if the original intent of the script was to tag the video-ingest devices (which tend to be "image") or cameras (which tend to be "camera").
- I don't know if or how there may be changes between individual devices across Windows or driver versions.
- This just lists a bunch of devices. This does **not** speak to whether those devices are certified, or if they'll show up as cameras **to the app**. Using a logical OR will make more things show up, **but may also make invalid things show up**. That said, there is no guarantee from the current logic that any listed devices are certified or will show up, _however_, the ingest devices tend to be classed as "imaging"

scanum:
@jambirk Hi James, following up on this. Thanks!

jambirk:
This is an inadvisable change to the recommended commands. Closing the PR.

PR #2221, opened 2019-05-08
Update network-requirements.md

Cloud-Writer:
Please add the following note to clarify the content. This request is via Content Idea Request 100702 and has been requested by Sr. EE Premal Gandhi. 
PSTN calls from the Skype for Business client usually use the G.711 codec, which requires a high bandwidth. If enough bandwidth is not available for that codec, then calls can fail with an error that resembles the following in the Media logs: Atleast one codec must be enabled, hr: c0042004. Media logs (.blog files) are encrypted and can be decoded only by Microsoft support personnel.


jambirk:
the markdown is malformed, I've sent corrections in email. Please re-submit.

PR #2310, opened 2019-05-21
Update manage-resource-accounts.md

hbunkradt:
Wrong comand should be  -OnpremPhoneNumber +14250000000


jambirk:
Confirmed the change in the command with the development team.

PR #2380, opened 2019-06-03
Update mobile-feature-comparison.md

PhillipGarding:
@jambirk - adding disclaimer text for the fact that user content is sent over the wire in APNS


PhillipGarding:
Thanks!

PR #2413, opened 2019-06-10
Service Number in Ressource Account

joaoalhinho:
In E5 there is already Phone System included.


scanum:
@jambirk Changes look correct. Could you please review and merge? Thanks!

jambirk:
this content is already under heavy rework. Will resolve this seperately.

PR #2465, opened 2019-06-14

carolynblanding:
@jambirk We had a reference in Step 6 for a next step that we removed. While writing internal docs, I realized we could use another step to verify upload of certificate so I added that. 

I added as verification step 1 (and bumped other steps up a number) to verify ability to change cloud voicemail greetings by SfB users with Exchange Hybrid Org.


scanum:
@jambirk

PR #2476, opened 2019-06-19
Update Known-issues.md

carolynblanding:
@LolaJacobsen 	@waseemhashem 	Added information that the department name will be missing when application endpoint is created with SfB Server 2015 which causes BVD provisioning failure as the BVD code relies on the department name. No code check in yet to resolve this.


LolaJacobsen:
@marclauinger - for your review

PR #2479, opened 2019-06-20
Update direct-routing-configure.md

NMURAV:


NMURAV:
Adding documentation for a new parameter in configuration command

PR #2503, opened 2019-06-23
chore: Add passing Markdownlint config

nschonni:
Suppressed all currntly failing rules.
Makes it less noisy editing with the MS Docs Authoring Pack


scanum:
@nenonix @JohanFreelancer9 Copy editing is needed for this PR. Thanks.

LolaJacobsen:
@dstrome could you look at this one?

scanum:

dstrome:
I'm not going to approve this change at this time. However, we'll review the markdownlint config syntax and, if it makes sense, will apply this broadly across the repos. Thanks for suggesting this @nschonni.

nschonni:
No problem, but note that it is already in azure-docs, dotnet, visualstudio, aspnet, etc...
Although only some teams actually laid out a preference of the rules vs. just adding the config like this and then cleaning up

dstrome:
Yup, understood. This is more a matter of ensuring that our repo configs are consistent. We have a few dozen repos and I want to avoid one-off configurations. This is something that makes sense to apply broadly so we'll investigate and roll out across our repos in a controlled way.

PR #2622, opened 2019-07-10

carolynblanding:
@jambirk This is somehow in the wrong tree. This should SfB Online. And we lost the SfB Server Oauth Integration with Exchange Online doc somewhere.


jambirk:
https://aka.ms/AdminDocsRequest is a more appropriate way to raise this issue, it requires updates to several files and changing file locations. 

@Carolynblanding we can deal with this using other methods, closing the PR

PR #2654, opened 2019-07-16
Update role-based-access-control-rbac.md

erwinbierens:
changing Lync to Skype for Business


MicrosoftHeidi:
Approving the reference change from Lync Server to Skype for Business Server.

PR #2683, opened 2019-07-21
Update location-based-routing-configure-network-settings.md

danny-levin:
use IPV6 as an example for site 2 instead of IPV4


scanum:
@nenonix @JohanFreelancer9 Copy editing is needed for this PR. Thanks.

PR #2691, opened 2019-07-22
ApplicationId missing from 'New-CSHybridApplicationEndpoint'

leeford:
The command to create the resource account is missing the application id, so you would not specify if it is for a call queue or auto attendant.


leeford:
Will send new PR.

PR #2729, opened 2019-07-29
Update to steps for enablement

SamCosby:
Need to update the steps for enablement. Currently, we have $RM but don't show customers who/what this is, so added a step to fix this.


LolaJacobsen:
@jambirk - for your review.

scanum:

PR #2797, opened 2019-08-06
Added details on provisioning.md

srividhc:


srividhc:
need to make more updates

PR #2858, opened 2019-08-20
Fix for malformed hardware requirements table

ctolkien:


scanum:

PR #2871, opened 2019-08-22
Clarificationons around taxes for porting numbers

savusca:
As per my knowledge, we do not request any taxes for porting numbers to Microsoft.
Can you please double-check this information with the appropriate PM and if validated, approve the article change?


CarolynRowe:
I have followed up with Sabin.  Adding pricing information to core content is not our policy.

PR #2872, opened 2019-08-22
Fix typo

robdy:
`Rata` should be `Data`. Obvious typo.


kenwith:
@robdy thanks for catch and fix, very much appreciated!

Issue #2876, opened 2019-08-22
Import PSSession missing


lcansby:
In the section "Create an account and set Exchange properties" Step 1 the powershell should end with importing the PSSession: 
Import-PSSession $sess -DisableNameChecking 

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: ee417633-3d52-646b-2ef0-94f71535db75
* Version Independent ID: 527bb7f0-b4a7-108c-6f86-8fadced63b86
* Content: [Deploy Microsoft Teams Rooms with Exchange Online](https://docs.microsoft.com/en-us/microsoftteams/room-systems/with-exchange-online#feedback)
* Content Source: [Teams/room-systems/with-exchange-online.md](https://github.com/MicrosoftDocs/OfficeDocs-SkypeForBusiness/blob/live/Teams/room-systems/with-exchange-online.md)
* Service: **msteams**
* GitHub Login: @LanaChin
* Microsoft Alias: **v-lanac**

scanum:
@lcansby Thank you for submitting feedback. We understand that this issue has been resolved.
Please feel free to re-open this issue if there is a specific area of the docs that we can improve or make better. Thank you.

PR #2901, opened 2019-08-27
Update with-exchange-online.md

lindspea:
https://github.com/MicrosoftDocs/OfficeDocs-SkypeForBusiness/issues/2876
Added Powershell step.

### DIFF
--- a/Teams/room-systems/with-exchange-online.md
+++ b/Teams/room-systems/with-exchange-online.md
@@ -42,6 +42,7 @@ Set-ExecutionPolicy Unrestricted
 $org='contoso.microsoft.com'
 $cred=Get-Credential $admin@$org
 $sess= New-PSSession -ConfigurationName Microsoft.Exchange -ConnectionUri https://outlook.office365.com/powershell-liveid/ -Credential $cred -Authentication Basic  -AllowRedirection
+Import-PSSession $Session -DisableNameChecking
 ```
 
 2. After establishing a session, you'll either create a new mailbox and enable it as a RoomMailboxAccount, or change the settings for an existing room mailbox. This will allow the account to authenticate into Microsoft Teams Rooms.


scanum:

LolaJacobsen:
@jambirk - please review

jambirk:
Original comment was "In the section "Create an account and set Exchange properties" Step 1 the powershell should end with importing the PSSession:
Import-PSSession $sess -DisableNameChecking"

@chrisda and @sohailta do you agree with this addition? I don't see any problem but I'd defer to either of your expertise.

chrisda:
The addition looks fine.

However, I would refrain from repeating the ExO PowerShell connection instructions in topics like this. Instead, I would simply link off to the official instructions: https://docs.microsoft.com/powershell/exchange/exchange-online/connect-to-exchange-online-powershell/connect-to-exchange-online-powershell.

One glaring issue I see is `Set-ExecutionPolicy Unrestricted`. This is a massive security hole that we should not espouse; `Set-ExecutionPolicy RemoteSigned` is what we require for ExO PowerShell, which is another reason to simply refer to the official connection instructions.

jambirk:
@lindspea please address the concerns ChrisDa raised

lindspea:
@jambirk @chrisda Thank you for the help. Just to confirm, although the addition is fine, it's recommended having it removed and rather adding a link in a note? @tseward fyi

chrisda:
The addition is fine. I would change the first step to something like:

1. Connect to Exchange Online PowerShell. For instructions, see [Connect to Exchange Online PowerShell](https://docs.microsoft.com/powershell/exchange/exchange-online/connect-to-exchange-online-powershell/connect-to-exchange-online-powershell).

In virtually all cases, that's the best move; there are too many variables with PowerShell these days (MFA? GCC High? etc.)

PR #2959, opened 2019-09-06
Update use-scom-management-pack.md

Ariantor:
Change from June to Sept.  The download shows 6/9 which is not correct. @jambirk please review and approve.


jambirk:
I was going off the info in the DLC information, but may have gotten this scrambled. Happy to approve and get it fixed.

PR #2988, opened 2019-09-11
Update Office-365-licensing.md

carolynblanding:
@LolaJacobsen , cc @bjwhalen 	

Not sure if I have the Note exactly correct, but I added this note " If Skype for Business Online is listed in apps for the selected license, do not remove it. If it's listed and is removed, Teams user setup will not complete successfully.".

Reason for the add is that we get many questions form Microsoft support engineers and customers about whether SfBO license is required for Teams. We also get case escalations where things stop working when customers remove SfBO app from licenses for users. 

If we could get this updated, that would go a long way to minimizing confusion and stopping case escalations.


scanum:

PR #3300, opened 2019-10-29
Update healthcare-templates.md

aaronlai425:
We are adding an additional template within the next week.


aaronlai425:
@tonysmit for your review

PR #3301, opened 2019-10-29
Updated Install Command

Rorymon:
Corrected from msiexec /i <path_to_msi> /l*v <install_logfile_name> ALLUSER=1 to
msiexec /i <path_to_msi> /l*v <install_logfile_name> ALLUSERS=1 (Added the S which is the valid MSI Property)


Rorymon:
Sorry, I was incorrect. I was using the parameter the wrong way.

PR #3302, opened 2019-10-29
Fixed typos

Rorymon:
Fixed two typos as didn't notice the second one when I submitted my first pull request.


Rorymon:
Sorry, I was incorrect. I was using the parameter the wrong way.

PR #3455, opened 2019-11-19
User can call someone in another company

markgort86:
It's only possible to call a guest user directly, when the user is signed in as a guest to the guest tenant. Otherwise you need to call the user as external contact which is possible.


scanum:

LolaJacobsen:
This isn't necessarily true. A guest user can call someone else regardless of whether they're signed in as a guest to the tenant IF it's PSTN or another voice call.

markgort86:
@LolaJacobsen then i don't understand the documentation about guest access and calling. If i'm logged in as an guest user (on the guest tenant) and want to call someone outside of that company within Microsoft Teams, there is no option to search users or adding new contacts. If this guest user has no teams license. How is it possible to call someone else (using a mobile phone is no option over here)?

markgort86:
@scanum is it possible to reopen this case? i'm still waiting on a reaction of @LolaJacobsen

PR #3489, opened 2019-11-22
Update call-park-and-retrieve.md

carolynblanding:
@CarolynRowe 

Feel free to refine my wording. Goal here is to clarify how users are grouped to the same call park group as it's not clear here that it's dependent on the call park policy assignment.


scanum:

PR #3508, opened 2019-11-26
Update direct-routing-migrating.md

carolynblanding:
@NMuravlyannikov , @CarolynRowe 
For Migrating from OPCH and CCE sections, not sure how the links to setting up OPCH and CCE provide information about migrating from these solutions because I didn't see any such information about migrating from these solutions in the referenced docs. 

Also for the recommendation to remove the CsVoiceRoutingPolicy assignment, we are showing a user level assignment which may or may not be effective depending on if a global voice routing policy is in place. Should we add a note regarding this recommending removing the global policy as well.


scanum:

scanum:
--

PR #3532, opened 2019-12-02
Update availability-in-india.md

msbemba:
made changes to audio conferencing availability


CarolynRowe:
Hello,
I checked with the responsible program manager and we will update this topic when given the OK.  The availability topics require Microsoft PM approval before they can be updated.

ipl-himanshu:
Any update on this? We have started using MS Teams for all our web meetings, but participants who are on the move can't Dial-in.

PR #3535, opened 2019-12-02
Updates to include package repositories and note about key and shipping.

kenwith:


Teams/get-clients.md
```diff
@@ -97,6 +97,11 @@ IT admins can use managed deployment of Teams to distribute the installation fil
 ### Linux
 
 Users will be able to install native Linux packages in `.deb` and `.rpm` formats.
+Installing the DEB or RPM package will automatically install the package repository
+- DEB `https://packages.microsoft.com/repos/ms-teams stable main`
+- RPM `https://packages.microsoft.com/yumrepos/ms-teams` 
+
+The signing key to enable auto-updating using the system's package manager can be found at: (https://packages.microsoft.com/keys/microsoft.asc). Microsoft Teams ships monthly and if the repository was installed correctly, then your system package manager should handle auto-updating in the same way as other packages on the system.
```
majo-javorka: signing key is also automatically installed, perhaps we should be explicit about that

PR #3538, opened 2019-12-02
Update limits-specifications-teams.md

msbemba:
Added limit on number of private channels.


LolaJacobsen:
Checking with PM before merging.

PR #3566, opened 2019-12-06
Cleanup docs for app developers

clearab:


clearab:
This PR removes outdated instructions for writing custom apps and instead adds pointers for the developer documentation.

PR #3577, opened 2019-12-10
Update to support Surface Hub meetings in GCC

gregbaribault:
As of November 1, Surface Hub was moved to a supported state for GCC tenants.


scanum:

Issue #3362, opened 2019-11-07
Is the first table correct? 


tophee:
I don't understand the last row in the first table (when to create a private channel) which states "Create a new team or create a new private channel in an existing team" even though there is no team that includes all members of the intended private group.

Shouldn't the recommendation for this last scenario be "Create a new team"? If that is too simplistic, I guess it would be "Add missing users as guests to the most relevant team and create a private channel or create a new team".

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: ce34b12b-aa50-608b-bbfa-a1fb4bb79f8d
* Version Independent ID: 9ddeb6f4-f30a-1e72-fc94-ae66def91f40
* Content: [Private channels in Microsoft Teams](https://docs.microsoft.com/en-us/MicrosoftTeams/private-channels#feedback)
* Content Source: [Teams/private-channels.md](https://github.com/MicrosoftDocs/OfficeDocs-SkypeForBusiness/blob/live/Teams/private-channels.md)
* Service: **msteams**
* GitHub Login: @LanaChin
* Microsoft Alias: **v-lanac**

scanum:
@tophee Thank you for submitting feedback and contributing to the docs. We are currently investigating this.

dariomws306:
Hi @tophee, thank you for your feedback.

I'm updating it to "Create a new private team or create a new private channel in an existing team after adding the missing users."

scanum:
@tophee Thank you for submitting feedback. We understand that this issue has been resolved.
Please feel free to re-open this issue if there is a specific area of the docs that we can improve or make better. Thank you.

PR #3658, opened 2019-12-17
Update private-channels.md

dariomws306:
https://github.com/MicrosoftDocs/OfficeDocs-SkypeForBusiness/issues/3362


LolaJacobsen:
Closing without merging. Per PM: While it is factually correct, seems like a bit more detail than necessary for a summary table.

scanum:
@dariomws306

PR #3790, opened 2020-01-16
Update migration-interop-guidance-for-teams-with-skype.md

carolynblanding:
Even though we state that Teams Phone System is not supported in Islands Mode, wanted to add clarity that DR numbers cannot be used if in Islands mode.


scanum:
@LanaChin PR has been copyedited and is ready for final review, could you please check and merge? Thanks!

PR #3843, opened 2020-01-24
Update turn-on-modern-auth.md

ndesaimsft:
Added another Note to explain that clients older than 2010 will not be able to login internally when using BlockWindowsAuthExternally


scanum:

PR #3846, opened 2020-01-24

gosepulv:
Proposing change since the tested behaviour verified is that even if the agent is offline or set to do not disturb, although the call does not notify the user, it is still routed to that user and fail to connect since the user is unavailable. There have been support requests about this matter. For the moment the call routing for call queues is not presence based. This can be the case in the future but there is no ETA for that to happen at this moment.

### DIFF
--- a/Teams/create-a-phone-system-call-queue.md
+++ b/Teams/create-a-phone-system-call-queue.md
@@ -197,9 +197,9 @@ Call agents selected must be:
 **Routing method** You can choose either **Attendant**, **Serial**, or **Round Robin** as the distribution method. All new and existing call queues have attendant routing selected by default. When attendant routing is used, the first call in the queue rings all call agents at the same time. The first call agent to pick up the call gets the call.
 
 - **Attendant routing** causes the first call in the queue to ring all call agents at the same time. The first call agent to pick up the call gets the call.
-- **Serial routing** incoming calls ring call agents one by one, from the beginning of the call agent list. Agents can't be ordered within the call agent list. If an agent dismisses or does not pick up a call, the call will ring the next agent and will try all agents until it is picked up or times out.
+- **Serial routing** incoming calls ring all call agents one by one, from the beginning of the call agent list. Agents can't be ordered within the call agent list. If an agent dismisses or does not pick up a call, the call will ring the next agent and will try all agents until it is picked up or times out.
   > [!NOTE]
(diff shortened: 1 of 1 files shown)


scanum:

Teams/create-a-phone-system-call-queue.md
```diff
@@ -197,9 +197,9 @@ Call agents selected must be:
...
+- **Serial routing** incoming calls ring all call agents one by one, from the beginning of the call agent list. Agents can't be ordered within the call agent list. If an agent dismisses or does not pick up a call, the call will ring the next agent and will try all agents until it is picked up or times out.
   > [!NOTE]
-  > Serial routing will skip agents who are **Offline**, have set their presence to **Do not Disturb**, or have **opted out** of getting calls from this queue.
+  > With Serial routing, for agents who are **Offline** or have set their presence to **Do not Disturb**, the call will be routed to those users and fail to connect unavailable user, the routing to the next agent in the agent list. This is not the case if the agent has **opted out** of getting calls from the call queue. To reduce the time interval the call routes to next agent in line, Agent Alert time can be decreased.
```
JohanFreelancer9:
```suggestion
  > With Serial routing, for agents who are **Offline** or have set their presence to **Do not Disturb**, the call will still be routed to those agents and and when failing to connect to the unavailable agents, will be routed to the next agent in the agent list. This is not the case if the agent has **opted out** of getting calls from the call queue. To reduce the time interval before the call routes to the next agent in line, Agent Alert time can be decreased.
```

AlexanderSaad: Exactly! Thanks for putting it that clear 😊

scanum:

Teams/create-a-phone-system-call-queue.md
AlexanderSaad: Hi @JohanFreelancer9 , do you have any idea when these changes will be applied to the official documentation ? Thank you !

AlexanderSaad: Hi @JohanFreelancer9 . It seems there is a grammar mistake here, as it does not make much sence:
_the call will be routed to those users and fail to connect unavailable user, the routing to the next agent in the agent list_
maybe that is more clear:
the call will be routed to those users and fail to connect **an** unavailable user, **after the specified agent alert time the call will be routed** to the next agent in the agent list.

JohanFreelancer9: @scanum It seems that my copy review was not been merged.

PR #3911, opened 2020-02-05
Update limits-specifications-teams.md

RuiPereiraTabares:
If you start typing a message that is reaching this limit you are not going to be able to send this chat message, there is no such limit for reactions or mentions in teams, there is a size limit because every reaction and mention will add size into the message.

### DIFF
--- a/Teams/limits-specifications-teams.md
+++ b/Teams/limits-specifications-teams.md
@@ -54,10 +54,12 @@ Teams chat works on a Microsoft Exchange backend, so Exchange messaging limits a
 |---------|---------|
 |Number of people in a private chat<sup>1</sup>  | 100    |
 |Number of file attachments<sup>2</sup>  |10     |
+|size of a team message<sup>3</sup>  |28 kb     |
 
 <sup>1</sup> If you have more than 20 people in a chat, the following chat features are turned off: Outlook automatic replies and Teams status messages; typing indicator; video and audio calling; sharing; read receipts.
 
 <sup>2</sup> If the number of attachments exceeds this limit, you'll see an error message.
+<sup>3</sup> If you start typing a message that is reaching this limit you are not going to be able to send this chat message, there is no such limit for reactions or mentions in teams, there is a size limit because every reaction and mention will add size into the message.
 
 ### Emailing a channel
 


LolaJacobsen:
@abgupta please review this change and let me know whether it's OK to merge (it needs editing, which I can do after you tell me whether it's a valid addition to this article).

LolaJacobsen:
Per abgupta, Nadia Fortini is the right PM to review this.

Nfortini:
What exactly are we trying to convey? The message is super confusing. It says that there is no such limit for reactions or mentions, but then it says "every reaction and mention will add size into the message."

For messages, it's a bit complicated. We have 28kb of space per post which includes the message itself (text, images, etc.), at mentions, reactions. This means that depending on 1) length of the message 2) number of reactions 3) number of at mentions and 4) number of connectors, the limit for the message length (same for, max # of reactions, max # of mentions) is not an exact limit, so we cannot say "this limit".  It's incorrect that there is no limit for reactions or mentions in teams, we found that the max for reactions is ~200 but I wouldn't communicate it since the exact number is dependent on the factors I mentioned above.

RuiPereiraTabares:
@Nfortini That is what i was trying to comunicate.
When we look in the raw server answer we will have the following:
{"errorCode":285,"message":"Failure due to: ImSizeTooBig."}
Reactions will increase the size of the message, so there is not such a thing as reactions limits, The message size limit is 28 KB which includes meta data, like reactions which can increase content size > to actual message content entered in the message.
When entering content in a message, 28 KB limit will be applied. If message content is 28 KB and someone tries to like it, then it will fail.
then the amount of reactions a message can have depends on the size of the body of the message.

RuiPereiraTabares:
customers and Support are believe that the channel message limit is the same as the teams message limit.. 
So maybe clarifiying that teams chat messages have a different limit will help Support and customers.

scanum:

Teams/limits-specifications-teams.md
```diff
@@ -54,10 +54,12 @@ Teams chat works on a Microsoft Exchange backend, so Exchange messaging limits a
 |---------|---------|
 |Number of people in a private chat<sup>1</sup>  | 100    |
 |Number of file attachments<sup>2</sup>  |10     |
+|size of a team message<sup>3</sup>  |28 kb     |
 
 <sup>1</sup> If you have more than 20 people in a chat, the following chat features are turned off: Outlook automatic replies and Teams status messages; typing indicator; video and audio calling; sharing; read receipts.
 
 <sup>2</sup> If the number of attachments exceeds this limit, you'll see an error message.
+<sup>3</sup> If you start typing a message that is reaching this limit you are not going to be able to send this chat message, there is no such limit for reactions or mentions in teams, there is a size limit because every reaction and mention will add size into the message.
```
JohanFreelancer9:
```suggestion
<sup>3</sup> If you type a message that will reach this limit, you will not be able to send the message. There is a size limit because every reaction and mention will add size into the message. There is no such limit for reactions or mentions in Teams.
```

Teams/limits-specifications-teams.md
```diff
@@ -54,10 +54,12 @@ Teams chat works on a Microsoft Exchange backend, so Exchange messaging limits a
 |---------|---------|
 |Number of people in a private chat<sup>1</sup>  | 100    |
 |Number of file attachments<sup>2</sup>  |10     |
+|size of a team message<sup>3</sup>  |28 kb     |
```
JohanFreelancer9:
```suggestion
|Size of a Teams message<sup>3</sup>  |28 kb     |
```

scanum:
@LolaJacobsen PR has been copyedited and is ready for final review, could you please check and merge? Thanks!

LolaJacobsen:
I'm closing this PR because I updated the article in a separate PR. THANK YOU to everyone who participated in this conversation.

PR #3965, opened 2020-02-14
What happened to the ability to do this on the admin center?

c9777777:
Hello, 

You used to be able to do this on the admin center on admin.microsoft.com, settings, services and addins.

If you look in this picture 

https://i.stack.imgur.com/plNQW.png

Ignore the sway box, but above it there is a Skype for business option to edit the external communications. 

How come that is no longer available on the admin.microsoft.com admin panel under services and addins?


LolaJacobsen:
Please enter this as an issue instead. It's not an appropriate addition to the article, so I'm closing this PR without merging.

PR #4250, opened 2020-03-23
Updated list of ELIN devices support

FilippSe:
Updated list of ELIN devices support and changed 'Yes' to tick mark


scanum:

PR #4466, opened 2020-04-08

carolynblanding:
Adding information for requirement to enable HMA for Exchange on-premises Server which is a step not handled by Oauth setup and is a hard requirement for Teams calendar delegation. 

@LolaJacobsen


carolynblanding:
Need to verify with PM and DEV.

PR #4567, opened 2020-04-16
Update Exchange-Teams-interact.md

carolynblanding:
Adding the additional steps required to allow for scheduling Teams calendar meetings on behalf of another user in Outlook. Reviewed these steps with Engineering and confirmed in the lab with in depth log analysis. The Teams scheduling application requires permissions to check the delegators mailbox for delegate permissions. The application is using the Skype for Business Online Application ID. To enable these permissions, the Exchange server needs to configure a Partner Application for the Skype Online Application ID with a linked account that has the Exchange UserApplication role assigned. 

With COVID 19 usage increases we've seen a large number of cases where delegation is not working due to these missing steps which were never documented. 

@LolaJacobsen @kenwith PM for this feature is ashwin.appiah@microsoft.com

PR #4599, opened 2020-04-19

michelderooij:
OAuth is only part of the requirement; the other requirement is that REST needs to be supported, and for that the mailbox needs to hosted on Exchange 2016 CU3+, per this article https://docs.microsoft.com/en-us/graph/hybrid-rest-support


dmstork:
I agree with Michel, I've come across several customers where they had coexistence with 2010/2013 and 2016. For them it wasn't explicit enough that the mailboxes that require the calender integration in Teams, have to be hosted on 2016CU3+ and that merely the presence of 2016CU3+ isn't enough.

dstrome:
Updated in #4567 ; closing this PR.

PR #4607, opened 2020-04-20
remove 50k port range

MartinRinas:
50k port range is not in use by the Teams service, would not recommend to mention it in this context at all. In fact only SfBO used this but we changed our guidance to no longer require this port range quite some time ago.


scanum:

MicrosoftHeidi:
We would be comfortable removing it if there was a certainty there would be no legacy use, please confirm the same (we can mark it Legacy if so).

MartinRinas:
yes, I am certain that the 50k port range is not used by the Teams service at all.

PR #4685, opened 2020-04-24
Add "share system audio" to the capabilities table

raramaia:
Share system audio will be generally available in Live Events on by Monday, 4/26. I added "Share System Audio on Windows" to the capability table, and note that it is only supported on Windows devices.


scanum:

scanum:
@LanaChin PR has been copyedited and is ready for final review, could you please check and merge? Thanks!

PR #4990, opened 2020-05-26
Update browser support to include video limits

emilykirby:
Added sup 3 to include that browsers only allow for a single incoming video stream from the active speaker. This is needed as folks are expecting to see 3x3 but that is not supported on the web (only mobile, Mac, and desktop)


scanum:

emilykirby:
This was already published. Neeraj Jain submitted separate PR. Thanks.

PR #4991, opened 2020-05-26
Update browser support to include video limits v2

nej-nej:
Update browser support to include video limits v2


scanum:

Issue #4753, opened 2020-04-30
Public Event


maggiori-git:
I think that it's important to write that "Anyone can attend a Teams meeting or live event for free - no license is required" is true only for PUBLIC live event. If we schedule a live event in the organization the attendee users need to have not only a generic Office365 license but almost a specific Teams license applied ( if live event is scheduled in Teams )

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 92fcb1f6-5cc2-f6bd-cfa9-11d811f1dbf1
* Version Independent ID: 9a571e6d-8cb3-9ff8-a5c2-62deab146264
* Content: [Admin quick start - Meetings and live events in Microsoft Teams - Microsoft Teams](https://docs.microsoft.com/en-us/microsoftteams/quick-start-meetings-live-events)
* Content Source: [Teams/quick-start-meetings-live-events.md](https://github.com/MicrosoftDocs/OfficeDocs-SkypeForBusiness/blob/live/Teams/quick-start-meetings-live-events.md)
* Service: **msteams**
* GitHub Login: @LolaJacobsen
* Microsoft Alias: **lolaj**

scanum:
@scanum  Thank you for submitting feedback and contributing to the docs. We are currently investigating this.

get-itips:
Hello @maggiori-git 
Thanks for your feedback. I will propose clarifying this article based on it, if it gets approval by the author, it will be published.

scanum:
@maggiori-git  Thank you for submitting feedback. We understand that this issue has been resolved.
Please feel free to re-open this issue if there is a specific area of the docs that we can improve or make better. Thank you.

PR #5077, opened 2020-06-05
Proposing clarifying to public live event

get-itips:
@LolaJacobsen please review if you agree to this change, this is based on 
https://support.microsoft.com/en-us/office/get-started-with-microsoft-teams-live-events-d077fec2-a058-483e-9ab5-1494afda578a?ui=en-us&rs=en-us&ad=us#bkmk_attendliveevents

> **Public events**
> 
> If an event is public, anyone who has the link can attend without logging in.
> 
> **Private events**
> 
> If attendance is restricted to your org or to specific people and groups, attendees will need to log in to join.
> 
> If the event is produced in Teams, they'll need a license that includes Teams. If it's produced externally, they'll need one that includes Microsoft Stream.

Related issue https://github.com/MicrosoftDocs/OfficeDocs-SkypeForBusiness/issues/4753

### DIFF
--- a/Teams/quick-start-meetings-live-events.md
+++ b/Teams/quick-start-meetings-live-events.md
@@ -30,7 +30,7 @@ There are 2 ways to meet in Microsoft Teams - meetings and live events. Use this
 
 ## Get licenses for meetings and live events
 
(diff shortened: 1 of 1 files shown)


scanum:

Issue #4494, opened 2020-04-09
Add a prerequisite - assign Microsoft as the audio conferencing provider


estiller:
I followed the steps on this guide; however, it resulted in various issues causing the assigned license not to work. After a lot of troubleshooting with Microsoft support, it turns out that the problem originated from the fact that the affected users had another audio conferencing provider set several years ago. This page should list setting Microsoft as an audio conferencing provider as a prerequisite (https://docs.microsoft.com/en-us/SkypeForBusiness/audio-conferencing-in-office-365/assign-microsoft-as-the-audio-conferencing-provider).

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 8ed33488-1a24-f443-dd56-c5327188b5a4
* Version Independent ID: 8da32180-396d-c7a8-6410-472beabff9ce
* Content: [Set up Audio Conferencing for Microsoft Teams - Microsoft Teams](https://docs.microsoft.com/en-US/microsoftteams/set-up-audio-conferencing-in-teams#feedback)
* Content Source: [Teams/set-up-audio-conferencing-in-teams.md](https://github.com/MicrosoftDocs/OfficeDocs-SkypeForBusiness/blob/live/Teams/set-up-audio-conferencing-in-teams.md)
* Service: **msteams**
* GitHub Login: @tonysmit
* Microsoft Alias: **tonysmit**

scanum:
@estiller  Thank you for submitting feedback and contributing to the docs. We are currently investigating this.

TokyoScarab:
Hi @estiller Thank you for raising this issue. I'll be looking into it!

scanum:
@estiller  Thank you for submitting feedback. We understand that this issue has been resolved.
Please feel free to re-open this issue if there is a specific area of the docs that we can improve or make better. Thank you.

PR #5101, opened 2020-06-09
Adding Prereq

TokyoScarab:
https://github.com/MicrosoftDocs/OfficeDocs-SkypeForBusiness/issues/4494

### DIFF
--- a/Teams/set-up-audio-conferencing-in-teams.md
+++ b/Teams/set-up-audio-conferencing-in-teams.md
@@ -33,7 +33,7 @@ You only need to set up Audio Conferencing for people who plan to schedule or le
 For frequently asked questions about Audio Conferencing, see [Audio Conferencing common questions](audio-conferencing-common-questions.md).
 
 > [!NOTE]
-> [!INCLUDE [updating-admin-interfaces](includes/updating-admin-interfaces.md)]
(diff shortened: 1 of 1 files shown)


scanum:

Teams/set-up-audio-conferencing-in-teams.md
```diff
@@ -34,7 +34,10 @@ For frequently asked questions about Audio Conferencing, see [Audio Conferencing
 
 > [!NOTE]
 > [!INCLUDE [updating-admin-interfaces](includes/updating-admin-interfaces.md)]
```
JohanFreelancer9, line 36:
```suggestion
> We are frequently updating how you can manage features, so the steps here might be a little different from what you see.
```

scanum:
@tonysmit PR has been copyedited and is ready for final review, could you please check and merge? Thanks!

tonysmit:
**@scanum - please don't merge this.** @TokyoScarab - what you are saying may be true for SFBO (which I don't believe it is either), but it's not correct for Teams. We don't support 3rd party AC providers in Teams. I just tested it; Assigned a license, then looked at the user's properties. There is no way to change the AC provider in Teams. This has been the case for several years now.

PR #5124, opened 2020-06-11
SEO edits for teams-in-hc

jdmartinez36:
SEO edits for the Get started with Teams for Healthcare organizations article.

The reviewer will want to carefully review the changes. The most common search terms for this article are "telehealth" and "telemedicine," neither of which fit well with the content. If telehealth is a broad term that encompasses all communication for healthcare, including internal, it would be fine; if it refers more to communication between patients and healthcare workers, this article doesn't seem to address that at all, since all of the content seems related to internal communications.

Telehealth is mentioned in the description and in a bulleted list, but nowhere else in the document. I've added a few mentions, but they might not make sense depending on the definition of these terms. The SEO expert recommended adding it to the metadata title, but it might create to much focus on that one feature.

The introduction implied this is a work in progress. If a section on telehealth is being added at some point, the edits to the body might not be necessary. 

Sorry for all the comments, this was a tough one to try and add these SEO terms to.

### DIFF
--- a/Teams/expand-teams-across-your-org/healthcare/teams-in-hc.md
+++ b/Teams/expand-teams-across-your-org/healthcare/teams-in-hc.md
@@ -1,5 +1,5 @@
 ---
-title: "Get started with Teams for Healthcare organizations"
+title: Get started with Teams telehealth for Healthcare organizations
 author: dstrome
 ms.author: dstrome
 manager: serdars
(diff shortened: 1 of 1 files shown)


jdmartinez36:
@DaniHalfin

tuco-ulf:

PR #5126, opened 2020-06-11
SEO edits for Shifts for Teams Landing Page

jdmartinez36:
Edited the description for SEO. The SEO expert also recommended adding a description under one of the headings to call out a search term (StaffHub alternative), but I felt it would disrupt the flow of the page too much. Or possibly break it. It also might confuse customers, since Shifts is not an alternative to StaffHub, but its replacement.

@DaniHalfin


tuco-ulf:

PR #5129, opened 2020-06-11
[DO NOT MERGE] SEO edits for Upgrade to Teams on-premises

jdmartinez36:
Updated the description and H1 per the SEO expert's recommendations. Added the "admin" part to the description since it's called out specifically in the H1.

The H1 change is based on "migrate" being the highest hit search term, but "upgrade" might be preferred as a marketing tool.  Another option might be to use "Upgrade and migrate...".

Removed a few extra spaces I noticed in the Overview.

@DaniHalfin

### DIFF
--- a/Teams/upgrade-to-Teams-on-prem-overview.md
+++ b/Teams/upgrade-to-Teams-on-prem-overview.md
@@ -8,26 +8,28 @@ ms.topic: article
 ms.service: msteams
 audience: admin
 ms.reviewer: bjwhalen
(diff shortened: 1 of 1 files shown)


tuco-ulf:

CarolynRowe:
I need to reject this one.  Marketing and the Upgrade product team are adamant that we use the term "Upgrade" and not "Migrate" or "Migration."  Please resubmit if there are other changes.  Going to close this one.  Thank you.

PR #5138, opened 2020-06-12
[DO NOT MERGE] SEO edits for Configure dynamic emergency calling

jdmartinez36:
Edited the title to add 911, did not add the - Microsoft Teams product name. I verified that the name is added automatically and appears in the search results.

Edited the description to add 911.

Added 911, e911, and Microsoft Teams e911 or emergency 911 to several parts of the document.

I think most or all of these changes will be rejected, as the document is worldwide. Only sections that specifically call out the United States might be able to keep the 911 mention.

@DaniHalfin

### DIFF
--- a/Teams/configure-dynamic-emergency-calling.md
+++ b/Teams/configure-dynamic-emergency-calling.md
@@ -1,5 +1,5 @@
 ---
-title: Configure dynamic emergency calling
+title: Configure dynamic emergency 911 calling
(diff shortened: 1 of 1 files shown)


tuco-ulf:

CarolynRowe:
Yes, need to reject this one.  Need to stay more generic--as 911 only applies to U.S. and could be misleading.

PR #5139, opened 2020-06-12
SEO edits for How Microsoft Teams uses memory

jdmartinez36:
Edited description and one image alt tag to include keywords.

The SEO expert recommended linking to an article that addresses Teams memory issues, but I was not able to find one.

@DaniHalfin


tuco-ulf:

PR #5149, opened 2020-06-15
Added Tendfor to the list of partners

dadavids:


dadavids:
Added Tendfor to the list of partners

PR #5316, opened 2020-07-08
Suggestion to add disclaimer

revaldiv:
At line 101 - I think we need to add some disclaimer about 3rd party apps.  Currently we do not have user specific reports about what apps a customer is using in Teams.  I am trying to help get info via Graph but in the Teams Admin Portal we do not have detailed reports about what apps a user has installed and is using.  Some are by Team so not sure how we would account for a team app.  Nonetheless I had a customer mention the following about the Karmabot app: I guess we can try reaching out to Karmabot for more info, we just find it concerning that Microsoft would allow all these third-party apps to perform functions within out Teams tenant and then have no insight into what the apps are doing or who is using them. That presents a huge compliance issue for us and if there is no way to get that information, we may need to disable all Teams apps across the tenant.


scanum:

PR #5350, opened 2020-07-15
Update pstn-usage-report.md

ckilaru-msft:
Updated language to reflect both in-direct calling plan providers.


ckilaru-msft:
need to make changes to 1 more file.

PR #5448, opened 2020-07-28
Align with race-insensitive language work

GossiTheDog:
As part of Race-Insensitive terms work across Microsoft.


scanum:

PR #5527, opened 2020-08-06
SPO and OD are not automatically IB enabled

anwara-ms:
SPO and OD are not automatically IB enabled, unless admin has requested IB to be turned on for SPO and OD.


robmazz:
Per conversation, closing and updating in private repo to fix spelling, grammar, and style issues.

PR #5613, opened 2020-08-20
Adjusting PS to install Public Preview version

get-itips:
Hello,

I am proposing to specify the Public Preview version as the current PS Line

``Install-Module MicrosoftTeams -AllowPrerelease``

**won't install the Public Preview but the GA version** and this is confusing users.

I know that hardcoding the version isn't the best, but again, the current instructions don't work as intended.

cc @brandber


brandber:
Thank you for raising!

Resolved in [PR #5616](https://github.com/MicrosoftDocs/OfficeDocs-SkypeForBusiness/pull/5616)

PR #5837, opened 2020-09-23
Identified new limit to private chat

kellybos:
Though a group chat can have up to 250 members, you can only add 200 members at one time. If you add more than 200 members at the creation of the chat you will get an error that the chat could not be sent, but this is due to the limit of adding members only 200 at a time. This was confirmed as a limitation by Teams engineering - https://portal.microsofticm.com/imp/v3/incidents/details/206239708/home.

Thanks,
Kelly

### DIFF
--- a/Teams/limits-specifications-teams.md
+++ b/Teams/limits-specifications-teams.md
@@ -60,16 +60,18 @@ Teams chat works on a Microsoft Exchange backend, so Exchange messaging limits a
 
 |Feature  | Maximum limit  |
 |---------|---------|
-|Number of people in a private chat<sup>1</sup>  | 250 |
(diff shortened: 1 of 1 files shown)


scanum:

JohanFreelancer9 (approved):

scanum:
@SerdarSoysal PR has been copyedited and is ready for final review, could you please check and merge? Thanks!

SerdarSoysal:
@scanum We may want to create a troubleshooting or known-issues article for this one. I'm following-up with our CSS org about this.

SerdarSoysal:
@scanum We are creating a KB for this. Will revise this PR accordingly when that is live.

PR #5838, opened 2020-09-23
Update stream-classification-in-call-quality-dashboard.md

sjaligam-MS:
For Satin code changes, Audio classifier needs to omit "Audio Degradation Avg"  and "Ratio Concealed Samples Avg" while determining PCR.


sjaligam-MS:
For Satin code changes, Audio classifier needs to omit "Audio Degradation Avg" and "Ratio Concealed Samples Avg" while determining PCR.

sjaligam-MS (commented):
@Priteshkanani @GageAmes : Please Review

GageAmes:
@sjaligam-MS Doesn't the classifier only change when the SATIN codec is in use? For codecs other than SATIN, is the old classifier supposed to be retained?

GageAmes (changes requested):
The previous classifier logic is still used for non-SATIN codecs, correct?

PR #6283, opened 2020-12-01
Clarification on language

PatmanW:


MSFTTracyP (changes requested):
Asking for clarification. It's not enough to say 'we don't use AES 256 across the service' for example. We need to know what is used, where something isn't used, and why. This is the all-up security guide for Teams.

MSFTTracyP (commented):
Spoke to Patrick and we'll go with this change.

PR #6333, opened 2020-12-10
Update plan-auto-attendant-call-queue.md

CLYVR:
Removing reference to "as they become available" in the general discription as this only happens if presence routing is enabled - ICM was opened due to misunderstanding of how/when calls routed to agents.


scanum:

JohanFreelancer9 (approved):

scanum:

PR #6352, opened 2020-12-14
Update to timing of limit increases + remove Adv Comms language

romanf01:
As prior deadline of Jan 1 2021 approaches for end of the Live Events limit increases, decision was made to extend limit increases until June 30, 2021. Also, value prop of Adv Comms SKU is still being landed, so removing that language. (Contacts: chtorok and zheh)


msdmaguire:
PubOps team, do not merge this until @romanf01 has gotten the go ahead. Ping dmaguire with questions.

msdmaguire:
Okay to merge. Merging now.

PR #6907, opened 2021-03-24
Update manage-tags.md

sandraluo:
Deleted the part about tagging by active shift coming soon now that it has publicly released.


scanum:

PR #6950, opened 2021-03-31
add de-identify settings for usage reports

ivlists:
Add setting information on de-identifying usage reports data


ivlists:
@cichur Can you please help merge these changes ?

Issue #5222, opened 2020-06-26
Dial-out behavior of organizer vs participant


mcantin:
Let's be more specific here. It has been confirmed by MSFT Support that disabling "dial-out" on an individual user (let's say Jake) will disable the meeting pane for everyone in a meeting Jake hosts. Whenever Jake clicks 'Phone Audio' before a meeting, he will receive an error saying 'Phone Audio isn't available. Unmute your mic for meeting audio'. Jake will be able to dial-out as a participant for meetings, hosted by individuals with dial-out enabled, through the meeting pane once inside the meeting, but not before.

This article doesn't clarify that, but in my opinion, it should.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: b759f354-bf74-3cff-0c96-3b28a27ee997
* Version Independent ID: fb2d8a68-f35a-bbee-a306-72ee20a450a3
* Content: [Set up the Call me feature for your users - Microsoft Teams](https://docs.microsoft.com/en-us/MicrosoftTeams/set-up-the-call-me-feature-for-your-users)
* Content Source: [Teams/set-up-the-call-me-feature-for-your-users.md](https://github.com/MicrosoftDocs/OfficeDocs-SkypeForBusiness/blob/live/Teams/set-up-the-call-me-feature-for-your-users.md)
* Service: **msteams**
* GitHub Login: @LanaChin
* Microsoft Alias: **v-lanac**

scanum:
@mcantin  Thank you for submitting feedback and contributing to the docs. We are currently investigating this.

dariomws:
Hi @mcantin, thank you for your feedback and help us to improve docs.microsoft.com.

I'm currently investigating this, I'll give you more information in the next couple of days.

dariomws:
Hi @mcantin, sorry for the delay.

I've verified, and if the meeting organizer is not enabled for dial-out, the **Phone audio** option won't be available for everyone in the "Choose your video and audio options" screen. So you won't be able even to get the **Call me** option.

If the meeting organizer is enabled for dial-out, the **Phone audio** option will be available for everyone in the "Choose your video and audio options" screen. So you can get the **Call me** option.

I'll update the article based on this information.

Thank you!

PR #6995, opened 2021-04-07
Update set-up-the-call-me-feature-for-your-users.md

dariomws:
Closes https://github.com/MicrosoftDocs/OfficeDocs-SkypeForBusiness/issues/5222.

I've verified, and if the meeting organizer is not enabled for dial-out, the Phone audio option won't be available for everyone in the "Choose your video and audio options" screen. So you won't be able even to get the Call me option.

If the meeting organizer is enabled for dial-out, the Phone audio option will be available for everyone in the "Choose your video and audio options" screen. So you can get the Call me option.

Also applied more updates to the article:
- Duplicate IMPORTANT note.
- Added a note that we usually see on these articles.
- Updated picture.
- Updated some steps.

Something very important to note is that in this way we are also disabling **Dial in manually** option. I think it's not expected and we should report it to the PG.

Thank you!


dariomws:
@cichur I think you can help to confirm and merge. Thank you!

cichur:
HI @dario-ms
Just a heads up, the image files can be no larger than 650px. I noticed that one file is 750px. Will you resize the images? Thank you, Cindy

dariopb:
@dariomws I think this was for you: "Just a heads up, the image files can be no larger than 650px. I noticed that one file is 750px. Will you resize the images? Thank you, Cindy"

cichur:
HI @dariomws, It looks like someone is committing new image files. I noticed that some of the new files are larger than 650px. I commented that the files need to be 650px or smaller. Cindy

dariomws:
@cichur done!

PR #7023, opened 2021-04-13
Change Localization Priority value

TinaMcN:
The localized topics get very high page views, let's change the value to "Priority" so that it will always be human translated.


scanum:

JohanFreelancer9 (approved):

PR #7024, opened 2021-04-13
Update localization priority value

TinaMcN:
The localized articles are getting high page views so let's upgrade the priority so it gets human translated.


scanum:

JohanFreelancer9 (approved):

PR #7029, opened 2021-04-14
Upgrade localization priority

TinaMcN:
This article gets high localization page views, suggest upgrading the priority so it gets human translated.


scanum:

JohanFreelancer9 (approved):

PR #7048, opened 2021-04-15
Add Meeting Scenario Dimension

rd-msft:
To enable identification of Microsoft Bookings' meeting scenario in CQD Dashboard through its own dimension filter


MicrosoftHeidi:
I would change the CY21Q2 language in the text, this may not be familiar to all users.

scanum:

PR #7234, opened 2021-05-17
Update plan-auto-attendant-call-queue.md

carolynblanding:
Fix missing requirement for phone number assignment


carolynblanding:
Need to make another adjustment

PR #7294, opened 2021-05-24
Update communicate-with-users-from-other-organizations.md

lambasu:
Update screenshot from External Access TAC page to include the following screenshot: Your Teams users can add apps when they host meetings or chats with other organizations, and they can use apps shared by external users when your users join meetings or chats hosted by other organizations. The data policies of the hosting user's organization, as well as the data sharing practices of the third-party apps shared by that user's organization, will be applied.


MikePlumleyMSFT:
Will update screenshot when ready via another PR.

PR #7359, opened 2021-06-04
Update configure-dynamic-emergency-calling.md

carolynblanding:
@CarolynRowe @roykuntz 
Adding information about how to define LIS Subnet


carolynblanding:
@CarolynRowe I'm going to resubmit with different network example to better align with cmdlet doc.

PR #7395, opened 2021-06-14
Update operator-connect-configure.md

livaht:
emergency addresses clarification (created in TAC, but changes can be made only by operator once associated with a number)
TeamsOnly clarification (user, not tenant based setting)


cazawideh:
@livaht Please review and let me know if this change looks correct. Thanks!

dariomws:
@livaht @cazawideh Are we sure that we need to create an emergency address and assign it to the user from the Admin Center? I'm working with NTT on a DRaaS project (not Operator Connect yet), and they assign a location (emergency address) to the DIDs they assign to the customer from their side. There is nothing that we need to do on the Teams Admin Center.

livaht:
@dariomws emergency address needs to be created in TAC, but is not assigned directly to a user but to a number by the operator when uploading the number to the tenant (and with the number to the user). Only operators can make changes to the emergency address assigned to the number. DR is different, the numbers are not uploaded to our inventory and also no TAC assignment is available. Hope this clarifies.

dariomws:
Thank you so much @livaht! It really clarifies for me.

PR #7475, opened 2021-06-29

shashank1320:
Use of Powershell for resource account related information in note section.
> You need to use the PowerShell cmdlets for Resource account number if not visible in drop down.

### DIFF
--- a/Teams/set-the-caller-id-for-a-user.md
+++ b/Teams/set-the-caller-id-for-a-user.md
@@ -33,6 +33,7 @@ By default, the following caller ID settings are **turned off**. This means that
     
   > [!NOTE]
   > If you want to use the *Service* parameter, you must specify a valid service number.
+  > You need to use the PowerShell cmdlets for Resource account number if not visible in drop down.
   
 - **Block outbound caller ID.** You can block the outgoing Caller ID from being sent on a user's outgoing PSTN calls. Doing this will block their phone number from being displayed on the phone of a person being called.
     


scanum:

Teams/set-the-caller-id-for-a-user.md
```diff
@@ -33,6 +33,7 @@ By default, the following caller ID settings are **turned off**. This means that
     
   > [!NOTE]
   > If you want to use the *Service* parameter, you must specify a valid service number.
+  > You need to use the PowerShell cmdlets for Resource account number if not visible in drop down.
```
JohanFreelancer9, line 36:
```suggestion
  > - If you want to use the *Service* parameter, you must specify a valid service number.
  > - You need to use the PowerShell cmdlets New-CsCallingLineIdentity or Set-CsCallingLineIdentity in the Teams PowerShell module 2.3.1 or later for the Resource account number if it is not visible in the dropdown.
```

JohanFreelancer9 (changes requested):
Copy review.

Teams/set-the-caller-id-for-a-user.md
JohanFreelancer9: @shashank1320 - I am really not sure about this - but your change seems incomplete. Please commit my suggested change if it is correct or edit your suggestion directly to provide complete information.

scanum:

Teams/set-the-caller-id-for-a-user.md
shashank1320: Updated with suggestion.

PR #7479, opened 2021-06-30
Deleted HK from MP locations

FilippSe:
HK DC has been deprecated, thus deleted it from the list


scanum:

PR #7511, opened 2021-07-07
Update sip-tester-powershell-script.md

ashhw:
Small grammar mistake, changing 'an SIP test' to 'a SIP test'.


scanum:

JohanFreelancer9 (commented):
@ashhw - "an SIP test" is actually correct. The indefinite article (an) is dependent on the pronunciation (sound) of the following term or word, not the first letter of the term or word (for example, it is “a Utopian idea,” but “an unfair world.”) So I would suggest that you please close this PR. Many thanks for the effort.

ashhw:
@JohanFreelancer9 - I stand corrected, I do apologise there, maybe I need to brush up on my grammar!

PR #7764, opened 2021-08-31
Update sfb-server-updates.md

Ariantor:
trying to resubmit this, this is the same as PR 7599.


Ariantor:
trying to resubmit this, this is the same as PR 7599. please let me know if this is the right way to do this.

PR #7785, opened 2021-09-03
Update cloud-recording.md

brguss:
Removed reference to the PowerShell policy attribute since it is not yet available.


brguss:
@naridi or @salarson do you have permission to add a label? If so, can you add the "sign-off" label to this PR?

Issue #7777, opened 2021-09-02
Planner not automatically created


LaurieKamp:
This page - wording and diagram seems to indicate that when a group is created, a Planner is also created.  I don’t believe that is true.  Creating a Planner for a Group is optional.  You have to manually go in and create it - either manually in Planner and attach it to the group - or by clicking Planner next to the group in the web version of Outlook. 

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 3d86301b-ce9c-388c-e563-9e5e5c27132e
* Version Independent ID: 07e4ae23-0966-e8f3-9b7c-4f6c02569eb0
* Content: [Microsoft 365 Groups and Microsoft Teams - Microsoft Teams](https://docs.microsoft.com/en-us/microsoftteams/office-365-groups)
* Content Source: [Teams/Office-365-groups.md](https://github.com/MicrosoftDocs/OfficeDocs-SkypeForBusiness/blob/live/Teams/Office-365-groups.md)
* Service: **msteams**
* GitHub Login: @MikePlumleyMSFT
* Microsoft Alias: **mikeplum**

scanum:
Thank you for submitting feedback and contributing to the docs. We are currently investigating this.

get-itips:
Hello @LaurieKamp Laurie, how are you? Hope you are doing good.
I could verify what you say here, I'll propose an improvement to the article.
Thank you

get-itips:
@scanum please close. thanks

PR #7792, opened 2021-09-06
Clarifying about planner

get-itips:
closes https://github.com/MicrosoftDocs/OfficeDocs-SkypeForBusiness/issues/7777

### DIFF
--- a/Teams/Office-365-groups.md
+++ b/Teams/Office-365-groups.md
@@ -29,7 +29,7 @@ By default, users in Microsoft 365 can create and manage groups. For more inform
 
 ## How Microsoft 365 Groups work with Teams
 
-When you create a team, a Microsoft 365 group is created to manage team membership. The group's related services, such as a SharePoint site, Power BI workspace, etc. are created at the same time.
+When you create a team, a Microsoft 365 group is created to manage team membership. The group's related services, such as a SharePoint site, Power BI workspace, etc. (Except Planner) are created at the same time.
 
 People who create teams can choose to use an existing Microsoft 365 group if they are an owner of that group. Each channel in the team has a separate folder in the document library. Creating folders directly in the document library does not create channels in the team.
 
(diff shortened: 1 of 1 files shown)


scanum:

Teams/Office-365-groups.md
```diff
@@ -29,7 +29,7 @@ By default, users in Microsoft 365 can create and manage groups. For more inform
 
 ## How Microsoft 365 Groups work with Teams
 
-When you create a team, a Microsoft 365 group is created to manage team membership. The group's related services, such as a SharePoint site, Power BI workspace, etc. are created at the same time.
+When you create a team, a Microsoft 365 group is created to manage team membership. The group's related services, such as a SharePoint site, Power BI workspace, etc. (Except Planner) are created at the same time.
```
JohanFreelancer9, line 32:
```suggestion
When you create a team, a Microsoft 365 group is created to manage team membership. The group's related services such as a SharePoint site and Power BI workspace (except Planner) are created at the same time.
```

JohanFreelancer9 (changes requested):
Copy review.

Teams/Office-365-groups.md
get-itips: I feel that if we remove the "," and replace with "and" we left out other services not explicitly mentioned.

JohanFreelancer9: Hi @get-itips. I suggested the change to remove the abbreviation "etc.", which, in accordance with the MS Style Guide, we should not use in docs. I have edited my suggestion, though. Please let me know what you think.

JohanFreelancer9 (approved):

scanum:

MikePlumleyMSFT:
@get-itips I think it's best not to call out Planner here. While the Planner instance isn't created until you access it from the team or group, the entry point is there and it auto-creates the first time you access it. You don't have to go to Planner and then connect to an existing group. I think calling it out as not being created will generate confusion.

get-itips:
> @get-itips I think it's best not to call out Planner here. While the Planner instance isn't created until you access it from the team or group, the entry point is there and it auto-creates the first time you access it. You don't have to go to Planner and then connect to an existing group. I think calling it out as not being created will generate confusion.

Hello Mike, no problem at all, thanks for your clarification.

PR #7924, opened 2021-09-29
Split initial note

YuriySamorodov:


scanum:

JohanFreelancer9 (approved):

scanum:
@cabailey PR has been copyedited and is ready for final review, could you please check and merge? Thanks!

cabailey:
I have reviewed & closing this PR without merging. The current formatting and order is the best user experience as well as consistent with other articles. We want to lead with the action and then an explanation, and it's always a best practice to avoid stacking callouts one on top of another. We also want to reserve Important callouts for things that are truly important, and it's really odd to start the page with this when all the articles in this doc set are for IT admins. 

If we lead with a "This page isn't for you" message in its own callout, end users who mistakenly end up here from search probably won't read any further and exit instead of selecting the link for them - that's why we put the action first, followed by an explanation within the same callout.  They are not isolated statements but cohesively joined by the reference to "these retention policies".

PR #8083, opened 2021-10-28
Update cloud-recording.md

lindspea:
https://github.com/MicrosoftDocs/OfficeDocs-SkypeForBusiness/issues/8072
Fixed heading

### DIFF
--- a/Teams/cloud-recording.md
+++ b/Teams/cloud-recording.md
@@ -213,7 +213,7 @@ If `Set-CsTeamsMeetingPolicy -ChannelRecordingDownload` is set to Block:
 
 - Channel members will have view only access without ability to download.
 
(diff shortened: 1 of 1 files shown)


lindspea:
@Markus-Hanisch on second glance I see that this table actually falls under the main heading 'Permissions and storage' so I have left the article as is and closed the PR. Thanks again for your valuable input.

Markus-Hanisch:
@lindspea 
Well, the issue was about the fact that the table does not belong to the previous subheading **Channel meetings** because it illustrates an overview on permissions and storage locations of the various meeting types.
So may the Table shall be moved or get its own subheading. 

And personally speaking, I like to have an extra subheading because I like link to that table directly from corporate documentation ;)

PR #8108, opened 2021-11-06
BulkAuthorUpdate

dariomws:
Update author from cichur to HowlinWolf-92.


dariomws:
@traya1, FYI.

Thanks!

Issue #8184, opened 2021-11-23
Which specific Power BI licenses needed


jiorblommaert:
Hello in the documentation it's mentioned the following "You can customize these query templates and distribute them to anyone in your organization who has a Power BI license and CQD admin permissions.". However which specific Power BI license is needed? Is it enough to use Power BI Free or does the user needs to have Power BI Pro?

scanum:
@jiorblommaert Thanks for pointing this out, could you please link the document you are referring to? Thanks.

jiorblommaert:
> @jiorblommaert Thanks for pointing this out, could you please link the document you are referring to? Thanks.
@scanum I'm referring towards following document: https://docs.microsoft.com/en-us/microsoftteams/cqd-power-bi-query-templates
In the beginning of the documentation the following sentence is mentioned "You can customize these query templates and distribute them to anyone in your organization who has a Power BI license and CQD admin permissions.". Some customers wants to know which specific Power BI license they need for this integration.

get-itips:
Hello @jiorblommaert I am looking into this.

get-itips:
Hello @jiorblommaert adjusting the article with the information I could get.

ParkesJ:
There are a lot of ways to license users for Power BI (see here: https://docs.microsoft.com/en-us/power-bi/consumer/end-user-license) and technically you could make do with a free Power BI license provided that the organization has a Premium subscription and a user with a Power BI Pro license assigned shares the report with you. Without that Premium subscription, a free license won’t work and you’d need Pro licenses.

Given all the possible licensing combinations that customers may have with respect to Power BI that would satisfy the requirements for sharing the CQD Power BI templates, it isn’t feasible to make prescriptive statements on Power BI licensing in CQD's documentation.

PR #8207, opened 2021-12-01
specifying license

get-itips:
closes https://github.com/MicrosoftDocs/OfficeDocs-SkypeForBusiness/issues/8184
confirmed with @ParkesJ Thanks!


ParkesJ (commented):
These changes -- as they are written -- are inaccurate. Power BI Pro is only a minimum license, it is not the required license (and technically a free license could work based on other Power BI licensing factors for a given organization). For this reason, we intentionally do not specify which license must be used, only that you need one of the available Power BI licenses.

get-itips:
> These changes -- as they are written -- are inaccurate. Power BI Pro is only a minimum license, it is not the required license. For this reason, we intentionally do not specify which license must be used, only that you need one of the available Power BI licenses.

Thanks for clarifying James

PR #8357, opened 2022-01-16
Adhere to current Debian recommendations for apt keys

jack1142:
The current recommendation for secure configuration apt keys is to use per-repo `signed-by=` option with the key stored in `/usr/share/keyrings`. This ensures that this key won't cause the system to accept signatures on *all* other repositories configured on the system that don't have a signed-by option (including the official Debian repositories).

From Debian's wiki:
> The key MUST NOT be placed in `/etc/apt/trusted.gpg.d` or loaded by `apt-key add`
https://wiki.debian.org/DebianRepository/UseThirdParty

This is just a documentation part, this would still need to be fixed in the package as `ms-teams` package unconditionally adds the key under `/etc/apt/trusted.gpg.d/microsoft.gpg` in `postinstall` script if it doesn't already exist.

### DIFF
--- a/Teams/get-clients.md
+++ b/Teams/get-clients.md
@@ -134,9 +134,9 @@ You can launch Teams via Activities or via Terminal by typing `teams`.
 Install manually on Debian and Ubuntu distributions:
 
 ```bash
-curl https://packages.microsoft.com/keys/microsoft.asc | sudo apt-key add -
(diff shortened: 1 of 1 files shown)


scanum:

JohanFreelancer9 (approved):

scanum:

Issue #8401, opened 2022-01-25
Serial Routing with Teams channels as Agent


javiernegro:
The documentation doesn't specify how Serial Routing works when the Agent is a teams channel.
Is there a way to set the order of the member of the group?
What criteria does it use to ring users?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: b755d9ae-13ac-019f-f5f4-a95b1eee2989
* Version Independent ID: f7afd8ad-e6e0-6c79-aed1-4ee87d48ba12
* Content: [Create a call queue in Microsoft Teams - Microsoft Teams](https://docs.microsoft.com/en-us/microsoftteams/create-a-phone-system-call-queue)
* Content Source: [Teams/create-a-phone-system-call-queue.md](https://github.com/MicrosoftDocs/OfficeDocs-SkypeForBusiness/blob/live/Teams/create-a-phone-system-call-queue.md)
* Service: **msteams**
* GitHub Login: @CarolynRowe
* Microsoft Alias: **crowe**

lindspea:
@javiernegro thank you for reaching out and helping improve Microsoft Docs. I've found some tech support articles that might be helpful. Please let me know. https://techcommunity.microsoft.com/t5/skype-for-business-preview/call-queue-serial-routing-agent-order/m-p/126117 and https://docs.microsoft.com/answers/questions/433192/order-agent-list-for-34serial-routing34-using-34vo.html

javiernegro:
Yes, I saw those. My comment  was to clarify and improve current docs.
Thanks!

PR #8456, opened 2022-02-03

lindspea:
https://github.com/MicrosoftDocs/OfficeDocs-SkypeForBusiness/issues/8401
Added links


scanum:

JohanFreelancer9 (approved):

scanum:

CarolynRowe:
the added links are not correct.  we should not be linking to a skype preview article.  also, i can't tell what the original request was.  what problem are we trying to fix here?  i'm going to close for now...  because i can't find the do not merge label and we should not merge the PR with the existing links.

javiernegro:
The problem is that the documentation is unclear about the behaviour of the product.
_The documentation doesn't specify how Serial Routing works when the Agent is a teams channel.
Is there a way to set the order of the member of the group?
What criteria does it use to ring users?_

PR #8517, opened 2022-02-15
Replace deprecated update guidance

flinchbot:
Remove incorrect reference that Windows updates happen during 2:00 maintenance Window. Update guidance to use the MTR-Update script and link to it.


cazawideh:
Making this change in private repo: https://github.com/MicrosoftDocs/OfficeDocs-SkypeForBusiness-pr/pull/8910

PR #8551, opened 2022-02-22
Update aa-cq-cqd-historical-reports.md

CLYVR:
Correcting description on one field.


CLYVR:
@CarolynRowe Please and thanks.

PR #8667, opened 2022-03-18
Update dimensions-and-measures-available-in-call-quality-dashboard.md

kellybos:
Identified another reason for null values for First and Second subnet. This information will not be reported when Teams web client is used.

https://portal.microsofticm.com/imp/v3/incidents/details/265325931/home


kellybos:
Provided evidence indicating this is not a valid reason for missing info. Closing PR.

PR #8718, opened 2022-04-01
Update get-clients.md

hemangupta1883:
In the following article we call out that Windows 10 LTSC does not Teams web but in this article we dont mention it which causes confusion to customers on support for Teams web on Windows 10 LTSC.
https://docs.microsoft.com/en-us/microsoftteams/hardware-requirements-for-the-teams-app


JohanFreelancer9 (approved):

scanum:

PR #8956, opened 2022-06-03
Remove Transport Relay Destination Port Note

gclark01:
Using the destination ports will not work for P2P Caller and Conf Callee Scenarios.   This might add confusion for customers who feel this would help them in their QoS deployments.


scanum:
@SerdarSoysal 

PR #9063, opened 2022-07-06
Inconsistency in the Engagement report doc

robertpp:
The Engagement report default value is Off. This is mentioned in the notes of [View and download meeting attendance reports in Teams](https://support.microsoft.com/en-us/office/view-and-download-meeting-attendance-reports-in-teams-ae7cf170-530c-47d3-84c1-3aedac74d310) and in [Microsoft Teams meeting attendance report](https://docs.microsoft.com/en-us/microsoftteams/teams-analytics-and-reports/meeting-attendance-report). The option is turned off when a new Teams meeting policy is created from Teams admin portal, but Enabled when a new Teams meeting policy is created in PowerShell.

image

image

### DIFF
--- a/Teams/meeting-policies-in-teams-general.md
+++ b/Teams/meeting-policies-in-teams-general.md
@@ -120,7 +120,7 @@ Keep in mind that after you set the default value, meeting organizers can still 
 
 This is a per-user policy. This setting controls whether meeting organizers can download the [meeting attendance report](teams-analytics-and-reports/meeting-attendance-report.md).
 
-This policy is on by default and allows your organizers to see who registered and attended the meetings and webinars they set up. To turn it off in the Teams admin center, go to **Meetings** > **Meeting policies**, and set the **Engagement report** setting to **Off**.
+This policy is off by default. If you turn this on, your organizers can see who registered and attended the meetings and webinars they set up. To turn it on in the Teams admin center, go to **Meetings** > **Meeting policies**, and set the **Engagement report** setting to **Turn on**.
 
 You can also edit an existing Teams meeting policy by using the [Set-CsTeamsMeetingPolicy](/powershell/module/skype/set-csteamsmeetingpolicy) cmdlet. Or, create a new Teams meeting policy by using the [New-CsTeamsMeetingPolicy](/powershell/module/skype/new-csteamsmeetingpolicy) cmdlet and assign it to users.
 


Teams/meeting-policies-in-teams-general.md
```diff
@@ -120,7 +120,7 @@ Keep in mind that after you set the default value, meeting organizers can still
 
 This is a per-user policy. This setting controls whether meeting organizers can download the [meeting attendance report](teams-analytics-and-reports/meeting-attendance-report.md).
 
-This policy is on by default and allows your organizers to see who registered and attended the meetings and webinars they set up. To turn it off in the Teams admin center, go to **Meetings** > **Meeting policies**, and set the **Engagement report** setting to **Off**.
+This policy is off by default. If you turn this on, your organizers can see who registered and attended the meetings and webinars they set up. To turn it on in the Teams admin center, go to **Meetings** > **Meeting policies**, and set the **Engagement report** setting to **Turn on**.
```
JohanFreelancer9, line 123:
```suggestion
This policy is off by default. If you turn it on, your organizers can see who registered and attended the meetings and webinars they set up. To turn it on in the Teams admin center, go to **Meetings** > **Meeting policies**, and set the **Engagement report** setting to **Turn on**.
```

JohanFreelancer9 (changes requested):
Copy review - @robertpp, kindly commit the minor suggested change. Many thanks.

robertpp:
Thank you for the review, @JohanFreelancer9. I'm not sure I'm doing it right, as I only have the 'Close pull request' button available.

robertpp:
Looks like I can't commit. I don't have write access to the repository.

JohanFreelancer9 (approved):
@robertpp, that is perfect, thank you.

CarolynRowe:
There is a problem with this PR.  The change is inconsistent with the paragraph following the change. 
@robertpp Can  you please clarify?

**Proposed change says the policy is off by default:**
This policy is off by default. If you turn this on, your organizers can see who registered and attended the meetings and webinars they set up. To turn it on in the Teams admin center, go to **Meetings** > **Meeting policies**, and set the **Engagement report** setting to **Turn on**.

But two paragraphs down still says the parameter is Enabled by default.   

**By default, the **AllowEngagementReport** parameter is set to **Enabled** in PowerShell.** To prevent a meeting organizer from downloading the meeting attendance report, set the **AllowEngagementReport** parameter to **Disabled**.

CarolynRowe:
If the following is true, then we need to clarify.  Because current change request doesn't clarify the following:  The option is turned off when a new Teams meeting policy is created from Teams admin portal, but Enabled when a new Teams meeting policy is created in PowerShell. I would also follow up with the PMs.  Is this the expected behavior?

robertpp:
Hello @CarolynRowe . This request was due to an incorrect understanding on my side based on incomplete information and erroneous assumptions. There are more details about this on [another request](https://github.com/MicrosoftDocs/office-docs-powershell/pull/9629). 

In summary, the policy is no longer off by default in the Global policy. According to the references mentioned in the first post and internal documentation, this was supposed to be off by default, due to privacy concerns. Something must have changed on how PG regards this feature.

There is also a dependency between the AllowEngagementReport value in the Global policy and the AllowEngagementReport in new meeting policies. All new policies follow the setting in the Global policy if not set. This is not documented anywhere and became the scope of the second GitHub item.

Finally, I made a confusion regarding the 'Turn on' visible in TAC when a new meeting policy is created. I interpreted it as an option to turn on the feature, meaning it is currently turned off, when it actually means the feature is turned on. Probably the best here would be to change 'Turn on' with 'Enabled' and 'Turn off' with 'Disabled', which would match the PS implementation and would prevent confusions, but this is not within the scope of this request.

To conclude, the actual purpose of this item remains alignment of the public documentation with respect to the actual default value of the policy. It turns out the needed change is to update the 2 references in the first post, which continue to state the default is off. At least one of them should probably mention also why such default is not regarded as a concern for privacy, since such default might raise a brow with our GDPR-compliant customers.

scanum:
@CarolynRowe

PR #9052, opened 2022-07-01
update members assign to tag

get-itips:
closes https://github.com/MicrosoftDocs/OfficeDocs-SkypeForBusiness/issues/9034

source https://github.com/MicrosoftDocs/OfficeDocs-SkypeForBusiness/pull/8867

PR #9082, opened 2022-07-12
Increase Members per Tag Limit

taisino:
Tag member limit was updated, and the following doc was already updated.
https://docs.microsoft.com/en-us/microsoftteams/manage-tags
https://github.com/MicrosoftDocs/OfficeDocs-SkypeForBusiness/pull/8867/commits/d8c6a8808c9971ffdf382ba3693dfea9d3940333
----------
A team can have up to 100 tags, up to 200 team members can be assigned to a tag, and up to 25 tags in the same team can be assigned to a single user.
----------

Updating this doc as well.


JohanFreelancer9 (commented):
Copy review - @taisino, this change is being dealt with in PR #9052 so this would be a duplicate and I suggest it can be closed. Many thanks.

PR #9266, opened 2022-08-31
Update the Conditional Access table for Phones

tsylvia88:
Update the Conditional Access table in the Teams Phones and Panels column.  Where it says "Grant access" it should say Supported in that column as well.


JohanFreelancer9 (approved):

scanum:

PR #9288, opened 2022-09-07
Update shared-channels.md

JasonLewisMSFT:
Removed as these two apps are not yet supported with shared channels.


JasonLewisMSFT (commented):
Please wait to approve, following up on if Calendar Pro really should be removed.

JohanFreelancer9 (approved):

Issue #9597, opened 2023-01-02
Some countries missing rom Zone 2 list


dipinn26592:
As per Below documents Pay as you Go is available for purchase in Mexico, Puerto Rico ad well as Slovenia, however these are not mentioned under zone 2 countries

https://learn.microsoft.com/en-us/microsoftteams/country-and-region-availability-for-audio-conferencing-and-calling-plans/availability-in-mexico
https://learn.microsoft.com/en-us/microsoftteams/country-and-region-availability-for-audio-conferencing-and-calling-plans/availability-in-slovenia
https://learn.microsoft.com/en-us/microsoftteams/country-and-region-availability-for-audio-conferencing-and-calling-plans/availability-in-puerto-rico

---
#### Document Details

⚠ *Do not edit this section. It is required for learn.microsoft.com ➟ GitHub issue linking.*

* ID: 243fa83d-b7bf-323a-a14a-026915d3d058
* Version Independent ID: fd4a310b-3144-4a8b-1bde-c2efd0c87377
* Content: [Calling Plans for Microsoft Teams - Microsoft Teams](https://learn.microsoft.com/en-us/microsoftteams/calling-plans-for-office-365)
* Content Source: [Teams/calling-plans-for-office-365.md](https://github.com/MicrosoftDocs/OfficeDocs-SkypeForBusiness/blob/live/Teams/calling-plans-for-office-365.md)
* Service: **msteams**
* GitHub Login: @CarolynRowe
* Microsoft Alias: **crowe**

dariomws:
@dipinn26592, thank you for your feedback.

We added the following note to the article:

> [!NOTE]
> Mexico has its own Pay-As-You-Go Calling Plan. It is not part of any zone. You must purchase the independent Pay-As-You-Go Calling Plan.

PR #9616, opened 2023-01-10
Add Mexico to zone 1 or 2 in calling-plans-for-office-365.md

dariomws:
Closes https://github.com/MicrosoftDocs/OfficeDocs-SkypeForBusiness/issues/9597.

### DIFF
--- a/Teams/calling-plans-for-office-365.md
+++ b/Teams/calling-plans-for-office-365.md
@@ -49,8 +49,11 @@ Here are the Calling Plans options:
   - **Pay-As-You-Go Calling Plan Zone-1**: Buy this for users in the United States and Puerto Rico, Canada, and the United Kingdom.
     - Zone 1 licenses aren't currently available for sale in the United States and Puerto Rico, but they can be provisioned to US and Puerto Rico users if bought in other countries.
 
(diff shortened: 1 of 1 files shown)


dariomws:
@CarolynRowe, could you help me to confirm if it's zone 2 for Mexico? Thanks!

dariomws:
@DaniEASmith, I think this question is better for you:

Could you help me to confirm if it's zone 2 for Mexico? Thanks!

JohanFreelancer9 (approved):

DaniEASmith:
Hey, @dariomws,

As it was explained to me by Chandra Kilaru, Mexico is its entirely own SKU. It isn't part of any zone. In this case, Mexico shouldn't be added to either zone list because if a customer wants Mexico Calling Plan PAYGo, they need to purchase the independent Mexico Calling Plan PAYGo license.

dariomws:
@DaniEASmith, thanks! I think we can add a note with this information. Do you agree? I can add it using this PR.

DaniEASmith:
> @DaniEASmith, thanks! I think we can add a note with this information. Do you agree? I can add it using this PR.

Yeah, I think that's a good idea.

dariomws:
@DaniEASmith, done!

DaniEASmith:
Looks good! I've merged.

PR #9653, opened 2023-01-23
Updated IP ranges to reflect correct state

FilippSe:
Updated IP ranges to reflect correct state


Adminiuga:
@FilippSe the `52.120.0.0/14` is incorrect range. There was an update and the correct range is `52.122.0.0/15`

PR #9666, opened 2023-01-25
Update view-only-meeting-experience.md

bwilsonms:
Changing from will to might because when the banner is thrown it just means the infrastructure is being spun up.


JohanFreelancer9 (approved):

scanum:
@mkbond007 

mkbond007:
@bwilsonms The "will" is explaining what the user experiences once the meeting has hit capacity and matches what the banner says. Is there a technical nuance that you want to call out?

bwilsonms:
> @bwilsonms The "will" is explaining what the user experiences once the meeting has hit capacity and matches what the banner says. Is there a technical nuance that you want to call out?

The banner can be sent to producers/co-producers as soon as the infrastructure spins up to support view-only, will seems to indicate it will happen immediately when this banner is shown, this is not the case.

scanum:
@mkbond007

PR #9726, opened 2023-02-14
Update set-up-webinars.md

jtl610:
Removing "Important" note about Microsoft List needing to be configured, which is no longer a requirement for the new webinar.


mkbond007:
No longer a requirement as mentioned, merging.

PR #9749, opened 2023-02-20
Fixed broken link

pvernocchi:


JohanFreelancer9 (approved):

scanum:
@DaniEASmith 

JohanFreelancer9:
> @scanum Does the URL need to have localization `en-us` string? Wouldn't that force readers of different languages to automatically view the English version rather their localized version of the page?

@DaniEASmith, I have checked and the link does not work without the localization `en-us` string (I am based in Vietnam). I can only assume the URL is not localized.

PR #9769, opened 2023-03-01
Correcting two measures

MattWade-MS:
These two measures are incorrectly stating they are stream measures, when in fact they are call based measures.  I've adjusted the definition language to reflect this difference.


JohanFreelancer9 (approved):

scanum:
@CarolynRowe 

PR #9827, opened 2023-03-16
Update room-operations.md

anwoodru:
Add updates for Teams Rooms on Window Client Cache Clear


anwoodru:
@microsoft-github-policy-service agree

PR #9951, opened 2023-04-18
Fix broken links, provide additional clarity

JustinLamMSFT:
Suggested updates:

In step 1
- I noticed that the Your products link is broken (at least for some users). It should point here: https://admin.microsoft.com/AdminPortal/Home#/subscriptions
- Also mentions "to check ... how many active users", but it's difficult to get that information from Your products. I suggest either that we tell them to go to Users > Active users: https://admin.microsoft.com/AdminPortal/Home#/users

In step 2
- Users > Active users is mentioned twice, but the link is broken (at least for some users). It should point here: https://admin.microsoft.com/AdminPortal/Home#/users


JustinLamMSFT:
#sign-off

JustinLamMSFT:
@microsoft-github-policy-service agree company="Microsoft"

Issue #9836, opened 2023-03-21
Update | reserved space for co-organizer, presenter and organizers


dipinn26592:
"The organizer will always be guaranteed space in the main meeting." - based on the latest update, the space is reserved for co-organizer, presenter and organizer,

FYI - @SamChung-MS 

---
#### Document Details

⚠ *Do not edit this section. It is required for learn.microsoft.com ➟ GitHub issue linking.*

* ID: 2bf705c2-1f24-bdb5-e712-e8408b941aca
* Version Independent ID: ddedd7e0-7d85-06e0-7f20-8b83b3dccec7
* Content: [View-only meeting experience - Microsoft Teams](https://learn.microsoft.com/en-us/microsoftteams/view-only-meeting-experience)
* Content Source: [Teams/view-only-meeting-experience.md](https://github.com/MicrosoftDocs/OfficeDocs-SkypeForBusiness/blob/live/Teams/view-only-meeting-experience.md)
* Service: **msteams**
* GitHub Login: @mkbond007
* Microsoft Alias: **mabond**

get-itips:
Thanks for your feedback

get-itips:
PR merged, closing

PR #10011, opened 2023-05-10
Add reserve for other roles

get-itips:
closes https://github.com/MicrosoftDocs/OfficeDocs-SkypeForBusiness/issues/9836


tiburd:
Info confirmed via email from Sam.

PR #10054, opened 2023-05-25
Update watermark-meeting-content-video.md

sadafbadar:
As per Security/privacy team we have to make sure that our wordings are very soft and we don't promise any security guarantee for feature Watermark. Removed the line such as "Meeting participants cannot turn off the watermark" and added few more context for IT admins regarding the licensing.


JohanFreelancer9 (changes requested):
Copy review - @sadafbadar, kindly commit the suggested changes. Many thanks.

sadafbadar:
All looks good!

PR #10190, opened 2023-07-17
Avatar profiles manageable by Global and User admins.

typride:
Updates to Global and user Admin wording, links to user docs, and reformatting/rewording.


typride:
@MikePlumleyMSFT would love a merge! Thank you!

Issue #10044, opened 2023-05-23
Request removal of 'Calendar Pro' from list of apps supported by Teams 'shared channel'


Xyvir:
Please see my last comment on the previously closed issue:

https://github.com/MicrosoftDocs/OfficeDocs-SkypeForBusiness/issues/9856

-Tyler

get-itips:
Hello @Xyvir 
I'm working on this and I'll get back.

PR #10204, opened 2023-07-20
Remove CalendarPro and OneNote

get-itips:
closes https://github.com/MicrosoftDocs/OfficeDocs-SkypeForBusiness/issues/10044

Checked with Jason Lewis

@tiburd


tiburd:
Confirmed that info matches email request. Thanks!

PR #10419, opened 2023-09-28
Update shared-calling-plan.md

justinmorris-MSFT:
Included (Optional) prefix on the PAYG Calling Plan/Communication Credits steps to avoid confusion for Direct Routing and Operator Connect customers that may interpret these steps as being required.


justinmorris-MSFT:
#sign-off

CarolynRowe:
Per review by @JensTrier, this PR is not correct.  Closing this PR.  We will work with Jens to make the meaning clearer--that it applies only to Calling Plan service numbers.

PR #10700, opened 2024-01-20
change words about receiveSurveysMode

SerinaJingyiLu:
We add a FF enableCQFWithReceiveSurveysMode. If FF is true, tenant admin could control CQF via the policy setting receiveSurveysMode


AndrewKlutz (approved):
approved change

PR #10702, opened 2024-01-24
Update teams-recording-policy.md

romanf01:
Add Communycarse as in-process (not yet certified)


MikePlumleyMSFT:
@2kiran can you review?

PR #10720, opened 2024-02-27
Add federation limitation

izzychun26:


MikePlumleyMSFT:
@izzychun26 can you clarify "This may cause unexpected domains on your allow list." It sounds like domains are getting added to the allow list, but I assume that's not the case.

izzychun26:
> @izzychun26 can you clarify "This may cause unexpected domains on your allow list." It sounds like domains are getting added to the allow list, but I assume that's not the case.

Hi Mike, yea that's not the case. I meant on your "external domain activity report" not "allow list". Let me know if this makes sense.

PR #10724, opened 2024-03-05
Update QoS-in-Teams.md

Petri-X:
There seems to be still missing the Teams V2 QoS advices for admins. Here is one proposal for helping them,


CarolynRowe:
This should have been fixed in the most recent update.  Closing this PR.  If there are still issues, please submit another PR.

PR #10726, opened 2024-03-06
Fix table layout

brien-c:
Fixed broken table layout for device models


MSTonySmith:
Fixed already in https://github.com/MicrosoftDocs/OfficeDocs-SkypeForBusiness/pull/10723

PR #10727, opened 2024-03-07
Update location-based-routing-india-plan.md

RasmusKoit:


CarolynRowe:
@RasmusKoit :  I need to check with the PMs on this.  We do not enable users for LBR for Operator Connect for India, so this addition would be technically incorrect.  For now, marking this as do not merge.

CarolynRowe:
Per @roykuntz, the requested addition is technically inaccurate.  Closing this PR.

PR #10732, opened 2024-03-12
Update shared-channels.md

yuokochi:
There are inconsistencies within the same document. Shared channels cannot be created by team members, but the table as being able to be created.

https://learn.microsoft.com/en-us/microsoftteams/shared-channels#shared-channel-creation
> Shared channel creation
> Only team owners can create a shared channel. Team members can't create them.


MikePlumleyMSFT:
@JasonLewisMSFT can you review?

JasonLewisMSFT:
The comment is correct, the table should be "No" for team member to create shared channels.

PR #10734, opened 2024-03-14
Update view-analytics.md

raul8a07:
Can the article include the note that this isn't available to users in the DoD tenant? I opened a case with Microsoft back in October and they mentioned that it wasn't available to users in the DoD and would like to have this article updated to share with users.


raul8a07:
Microsoft Case # for reference: 2310170040009546

PR #10744, opened 2024-03-28
Update manage-the-usage-of-a-phone-number.md

ChristinRichter:
Adding additional information as the page is very US centric currently. Adding information about the restriction of self-conversion will help customers to get to the goal faster


ChristinRichter:
tried to add label "Sign Off" but this is not available for me (adding any label that is)

ChristinRichter:
I cannot add any label, can this be considered "Sign Off", please

PR #10752, opened 2024-04-03
Update bring-your-own-device.md

vexir:
Added verified device list.


MSTonySmith:
@vexir - Because you are an FTE, this must be moved over to the private repo and assigned to me. Closing.

PR #10763, opened 2024-04-18
Update cloud-intelliframe.md

techbrett:
PTZ Pro 2 Device is no longer an Android Teams Rooms certified device per this document: https://learn.microsoft.com/en-us/microsoftteams/rooms/certified-hardware?tabs=Devices as it no longer has a minimum certified firmware version.


techbrett (commented):
Found that this device is no longer certified for Teams Rooms here: https://learn.microsoft.com/en-us/microsoftteams/rooms/certified-hardware?tabs=Devices (scroll down to PTZ Pro 2 - no Minimum firmware version).

PR #10787, opened 2024-05-16
we do not want to encourage chat using guest access. Further, the very next note talks about the right way of chatting and the word chat in this paragraph is likely to create confusion to readers.

imkundan:


jacktremper:
@MicrosoftHeidi are you able to review and merge as needed?

PR #10800, opened 2024-05-31
Update new-teams-deploy-with-m365apps.md

SAICHendrix:
This information when new Teams will be included in Semi-Annual Enterprise Channel (SAEC) is incorrect based on testing. SAEC still includes classic Teams as of a week ago. SAEC (Preview) does include new Teams. Please validate finding and if needed please update this with the correct date at which SAEC is expected to receive new Teams bundled with it.


MicrosoftHeidi:
THIS IS NOT TO BE PUBLISHED.

PR #10802, opened 2024-06-13
Update sip-gateway-configure.md

shanehoey:
Added Troubleshooting steps for 911 calls not working on SIP Compatible Phones. 
).


shanehoey:
I did a bit more testing I think it can be reworded just to  mention the Emergency Call Routing Policy

> [!NOTE]
> **Troubleshooting Emergency Calls** : Dialing 911 from a SIP Gateway Connected phones may result in the Call unable to reach 911. If dialing 911 on a SIP Gateway Connected phone fails, and dialing 911 on other Teams clients is successful, check the Emergency Call Routing Policies. SIP Gateway phones exhibiting this behaviour require a Emergency Call Routing Policy for 911 calls. For more information about emergency calling, see [Plan and manage emergency calling](/microsoftteams/what-are-emergency-locations-addresses-and-call-routing).

PR #10807, opened 2024-06-21
Update teams-classic-client-uninstall.md

wayvad:
Update the admin guide to point to the correct page.


wayvad:
I would like to point to a specific section on this page (so with anchor): https://learn.microsoft.com/en-us/microsoftteams/new-teams-bulk-install-client#remove-classic-teams-for-all-users

Is it possible to do so? Can I add "#remove-classic-teams-for-all-users" to the .md relative reference?

PR #10855, opened 2024-08-30
Update tmr-meeting-recording-change.md

lesahill:
I think that this copy is confusing. I edited some language to make it more concise, but I think the first paragraph in the 'Meetings and events' section could benefit from further simplification.

I am editing this documentation because I leverage it for self-help solutions, which are provided to admins during the support ticket creation flow. Simple language is important to help admins understand their options, and therefore, empower them to administer Teams by themselves. This helps deflect support volume, which is a cost to Microsoft.


wlibebe:
hi @lesahill thank you for these suggestions. i've added them to another active PR that'll be published today.

PR #10866, opened 2024-09-30
Switched order of PowerShell instructions

temanteleary:
Updated article to switch order around and have PowerShell instructions appear at the bottom instead of at the top.


CarolynRowe:
The PowerShell instructions were moved to the wrong section of the doc.  Instructions for assigning were moved to the changing section, I will create a separate PR to fix.

temanteleary:
> @temanteleary  Please work in the private repo, not the public. Contact me if that's confusing.

@garycentric Will do. I figured out what happened so it doesn't happen again. Thanks.

PR #10879, opened 2024-10-22
Update call-sharing-and-group-call-pickup.md

emilykirby:
Adding limitation on ring order, and cleaning up the limitations

QUESTION: //what does characters signify? I am not sure that this is referring to


mkbond007:
Closing this PR, made updates in separate PR.

PR #10896, opened 2024-11-21
Update manage-queues-app.md

emilykirby:
Updating to say that VDI is now supported


mkbond007:
Already done in another PR in the private repo

PR #10924, opened 2025-01-31
Update block-incoming-chats.md

hgautam26:
Removing the presence status text.


hgautam26:
Removed the text that states we don't show presence status.

PR #10930, opened 2025-02-07
Update anonymous-users-in-meetings.md

Yebbenbe:
Corrected org-wide vs per-organizer anon user access, these were backwards. You can't apply meeting SETTINGS to people, you do this with meeting POLICIES. I thought maybe I was just misunderstanding, so I asked co-pilot and it just got stuck in a loop of contradictions. So I'm assuming it's wrong? It's still unclear how the global policy is involved in all of this, is it not just a redundant form of the Meeting Settings? I think the global policy is being phased out, IIRC. If so, maybe a note should be added there


Yebbenbe:
@microsoft-github-policy-service agree

